PR #10, opened 2016-01-04
Added commas to protocols.csv

wanderer:


lgierth:
Why the need for commas, aren't tabs fine?

wanderer:
@lgierth because github make it look cool and searchable if you do https://github.com/wanderer/multiaddr/blob/master/protocols.csv 

Let me know which way you prefer!

bcavileer:
+1
I guess I don't need to submit my pull request ;)

jbenet:
sure sgtm

jbenet:
hopefully nothing expecting tabs broke

PR #11, opened 2016-01-04
added 420 rlpx

wanderer:
I have no idea how the numbers are chosen. But 420 is a great number


lgierth:
Descriptive commit messages are a thing btw ;)

Is rlpx this? https://github.com/ethereum/devp2p/blob/master/rlpx.md
What are the plans with this protocol in multiaddr? I might have missed out-of-band discussions.

protocols.csv
```diff
@@ -1,12 +1,13 @@
...
+code,	size,	name
+4,	32,	ip4
+6,	16,	tcp
+17,	16,	udp
+33,	16,	dccp
+41,	128,	ip6
+132,	16,	sctp
+301,	0,	udt
+302,	0,	utp
+420, V, rlpx
```
lgierth: delimiting with spaces here, should be tabs

wanderer: @lgierth fixed

wanderer:
>Is rlpx this? https://github.com/ethereum/devp2p/blob/master/rlpx.md

yep that is RPLx.

>What are the plans with this protocol in multiaddr? 

I'm trying to make RPLx compatible with @diasdavid 's [interface-connection](https://github.com/diasdavid/interface-connection]

lgierth:
Wow -- sorry this has been sitting here for so long. Do you still have a need for `/rplx` multiaddrs? I kinda gave 420 to `/p2p` in #27 (which will replace `/ipfs`), but something like 460 should be fine. What would these roughly look like?

PR #15, opened 2016-03-05
Add py-multiaddr

sbuss:
I'm not sure if you want alpha-level implementations on the list. I just wrote a python implementation, trying to match your API as closely as possible. It could use a few more tests (and maybe some vectors we can use across languages).

I still need to package it and release it to pypi, but that will probably happen tonight or tomorrow.


jbenet:
Yep, this is great. i'll add it manually as now the PR will conflict

PR #16, opened 2016-03-29
Added a link to the Haskell implementation

basile-henry:
I have implemented multiaddr in Haskell looking at the Go implementation for the details. I believe it is stable enough, but I will look into getting some peer review from the Haskell community.


jbenet:
Fantastic, thanks!

basile-henry:
You're welcome, great protocol by the way! :+1:

PR #23, opened 2016-09-26
/ipfs code mismatch

RichardLitt:
This is taken from https://github.com/multiformats/js-multiaddr/blob/master/src/protocols.csv#L10. I am not sure which is more accurate, this or that file. Any idea where to find the source of truth for this address?

Perhaps we should collapse the two protocols list, or import multiaddr's list into js-multiaddr directly.

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -7,7 +7,7 @@ code,	size,	name
 132,	16,	sctp
 301,	0,	udt
 302,	0,	utp
-421,	V,	ipfs
+42,	V,	ipfs
 480,	0,	http
 443,	0,	https
 444,	96,	onion


lgierth:
@dignifiedquire go-ipfs uses 421 and js-ipfs uses 42 -- this looks fishy to me, could you confirm it's a bug (or not a bug)?

dignifiedquire:
@diasdavid thoughts on this?

diasdavid:
That is just the `.csv` table that is wrong, the actual table used (in the .js file) has 421 

https://github.com/multiformats/js-multiaddr/blob/master/src/protocols.js#L41

🐟 🐟 🐟 🐟 🐟

lgierth:
Mind removing that table in js-multiaddr?

diasdavid:
done https://github.com/multiformats/js-multiaddr/commit/b8edb7e174cc62f57028ed044163239b8abab896

dignifiedquire:
Thank you @diasdavid

PR #31, opened 2016-12-24
Change /ip6 to variable-length

lgierth:
IPv6 addresses have a maximum size, but they can be shortened under certain circumstances, e.g. null-bytes can be omitted.

I know at least js-multiaddr and go-multiaddr permit variable-length /ip6 addresses anyhow, and also I'm not sure anything is actually paying attention to the sizes defined in protocols.csv :)

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -3,7 +3,7 @@ code,	size,	name
 6,	16,	tcp
 17,	16,	udp
 33,	16,	dccp
-41,	128,	ip6
+41,	V,	ip6
 132,	16,	sctp
 301,	0,	udt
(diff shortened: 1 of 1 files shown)


Kubuxu:
The length in this document is length of byte encoded address. I think it should stay 16 and we should harden js and go-multiaddr to always require 16 bytes of IPv6 address.

The shorter from (with missing bytes replaced by 0s) hardly ever will be valid reachable address.

lgierth:
That makes sense, thanks @Kubuxu. I'll leave this open as a TODO for the upcoming /ip specs.

Issue #22, opened 2016-09-16
/dns in multiaddrs


jbenet:
There's been some conversation on how to handle DNS, in these issues and others:
- https://github.com/multiformats/go-multiaddr/issues/7
- https://github.com/multiformats/go-multiaddr/pull/26

Finally looking back at this-- sorry for the delay everyone -- @sivachandran @lgierth @Gaboose and more. I am going to collect the options here and make a decision in coming weeks.
### Proposals so far

So far, the proposals are:
#### Proposal 1: `/dns/<domain>/<proto>`

``` sh
# format
/dns/<domain>/<proto> 

# example
/dns/example.com/ip4  --resolves-to-->  /ip4/1.2.3.4
```
#### Proposal 2: `/dns/<domain>/<record-type>`

``` sh
# format
/dns/<domain>/<record-type> 

# possibilities / examples 
/dns/example.com/A     --resolves-to-->  /ip4/1.2.3.4
/dns/example.com/AAAA  --resolves-to-->  /ip6/1::1
/dns/example.com/CNAME --resolves-to-->  /dns/another.com/*
/dns/example.com/MX    --resolves-to-->  /dns/mail.another.com/* 
/dns/example.com/A     --resolves-to-->  /dns/another.com/A  (through a CNAME)
/dns/example.com/TXT   --resolves-to-->  /ipfs/QmZS2PeM6rtSJXPPSBiqyNRss3M1JpuNAkkQmzEWR7aw1w  (through a dnslink)
```
### Other Questions
- could we support both proposals?
- are there other proposals?
- can we come up with a long list of all the concrete example resolutions we want? (engineer to the use case)
- maybe (for CNAME, MX, etc) we can return a multiaddr that is "[either ip4 or ip6](https://github.com/multiformats/multiaddr/issues/21)" or a "[combination](https://github.com/multiformats/multiaddr/issues/20)"?
- what would just `/dns/example.com` mean? a list or combination? (`1 -> many`)

ProgVal:
Other proposal: have both a modified version of proposal 1 (let's call it 1b) and proposal 2.

*proposal 1b:* `/resolution/name/proto`

`/resolution/name/proto` could be resolved to any resolution protocol, eg. DNS.

For instance, `/resolution/example.com/ip6    -->    /dns/example.com/AAAA    -->    /ip6/1::1`

ProgVal:
[...]

PR #34, opened 2017-01-17
Add /dns, /wss, /relay, /p2p protocol codes

lgierth:
Also add a fourth column, for comments.

- We'll be using the experimental /dns-exp protocol for now, as per #22. Once we're sure it's right, we'll make it into /dns.
- Same procedure for /relay, where in addition to the semantics, we're not even sure about the name.
- /wss on the other hand is pretty straightforward.

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -14,15 +14,14 @@ code,	size,	name,	comment
 420,	V,	p2p,	preferred over /ipfs
 421,	V,	ipfs,	equal to /p2p
 444,	96,	onion
-450,	V,	relay,	reserved for near-future use
 480,	0,	http
 443,	0,	https
 477,	0,	ws
 478,	0,	wss
 275,	0,	libp2p-webrtc-star
 276,	0,	libp2p-webrtc-direct
+290,	V,	libp2p-circuit-relay
 
 16383,	V,	exp-dns,	experimental /dns
 16382,	V,	exp-dns4,	experimental /dns4
 16381,	V,	exp-dns6,	experimental /dns6
-16380,	V,	exp-relay,	experimental /relay
--- a/protocols.csv
+++ b/protocols.csv
@@ -1,9 +1,11 @@
-code,	size,	name
+code,	size,	name, comment
 4,	32,	ip4
 6,	16,	tcp
 17,	16,	udp
 33,	16,	dccp
 41,	128,	ip6
+53,  V, dns, reserved for future use
+531, V, dns-exp, experimental /dns
 132,	16,	sctp
 301,	0,	udt
 302,	0,	utp
@@ -11,5 +13,8 @@ code,	size,	name
 480,	0,	http
 443,	0,	https
 444,	96,	onion
+450,  V,  relay, reserver for future use (preliminary name)
+451,  V,  relay-exp, experimental /relay
 477, 0, ws
+478, 0, wss
 275, 0, libp2p-webrtc-star
--- a/protocols.csv
+++ b/protocols.csv
@@ -1,20 +1,27 @@
-code,	size,	name, comment
+code,	size,	name,	comment
 4,	32,	ip4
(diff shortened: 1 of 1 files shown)


lgierth:
Hey, updated this -- this PR now covers:
- /p2p
- /wss
- experimental /dns, /dns4, /dns6
- experimental /relay
- /unix which was already in go-multiaddr

protocols.csv
```diff
@@ -1,15 +1,27 @@
...
 443,	0,	https
-444,	96,	onion
-477, 0, ws
-275, 0, libp2p-webrtc-star
\ No newline at end of file
+477,	0,	ws
+478,	0,	wss
+275,	0,	libp2p-webrtc-star
+
+16383,	V,	exp-dns,	experimental /dns
```
lgierth: This is based on the highest value of a 2-byte uvarint: 2^14 - 1 = 11 1111 1111 1110 = 16383

protocols.csv
```diff
@@ -1,15 +1,27 @@
...
 443,	0,	https
-444,	96,	onion
-477, 0, ws
-275, 0, libp2p-webrtc-star
\ No newline at end of file
+477,	0,	ws
+478,	0,	wss
+275,	0,	libp2p-webrtc-star
+276,	0,	libp2p-webrtc-direct
+290,	V,	libp2p-relay
```
diasdavid: can it be `libp2p-circuit-relay`, pretty please? :)

protocols.csv
```diff
@@ -1,15 +1,27 @@
...
+
+16383,	V,	exp-dns,	experimental /dns
+16382,	V,	exp-dns4,	experimental /dns4
+16381,	V,	exp-dns6,	experimental /dns6
```
diasdavid: Do we need it to be experimental? Isn't it going to be more of an 'experimental implementation', than a 'experimental addr'?

protocols.csv
diasdavid: @lgierth

lgierth: yep updated! :+1:

protocols.csv
lgierth: > Do we need it to be experimental? Isn't it going to be more of an 'experimental implementation', than a 'experimental addr'?

It's probably experimental on both address semantics and implementation.

The practical reason is I'm not set on whether we should do only /dns, or only /dns[4,6], or both, and would like to implement both before settling on an option.

The process reason is I'm pretty confident moving fast with protocols prefixed with /libp2p- and protocols that don't have values (like /ws), because they're likely easy to change or abandon. I feel like the more essential protocols like /ip4 and /dns need more care.

diasdavid: Understood :)

protocols.csv
```diff
@@ -1,15 +1,27 @@
...
 443,	0,	https
-444,	96,	onion
-477, 0, ws
-275, 0, libp2p-webrtc-star
\ No newline at end of file
+477,	0,	ws
+478,	0,	wss
+275,	0,	libp2p-webrtc-star
+276,	0,	libp2p-webrtc-direct
+290,	V,	libp2p-circuit-relay
```
diasdavid: Is relay a multiaddr? Should be a 'mount' protocol, no?

lgierth: The way I wanna do it (haven't gotten much feedback) is for it to be a mount/swarm protocol and transport at the same time. The former for speaking relay with other nodes, the latter to actually send stuff through it.

diasdavid: Oh I see. We will need to update it also here: https://github.com/libp2p/specs/blob/master/7-properties.md#757-protocol-multicodecs (that table is from April 2016)

You mention not getting much feedback on your proposal, where can I read the circuit relay spec and provide my review?

lgierth: Yep will update the protocols table.

Cool to merge?

PR #51, opened 2017-09-10
fix QUIC

marten-seemann:
The QUIC code point was changed in https://github.com/multiformats/multicodec/pull/57.


diasdavid:
thanks @marten-seemann :)

PR #57, opened 2017-11-27
Add /p2p-websocket-star

lgierth:
cc @mkg20001 @diasdavid


protocols.csv
```diff
@@ -21,4 +21,5 @@ code,	size,	name,	comment
 478,	0,	wss,
 275,	0,	libp2p-webrtc-star,
 276,	0,	libp2p-webrtc-direct,
+479,	0,	p2p-websocket-star
```
diasdavid: Thanks @lgierth! 

Seems that `libp2p-webrtc-star` and `libp2p-webrtc-direct` never got updated here. Mind updating them to their `p2p-` versions as well?

PR #64, opened 2018-05-24
Make /p2p into an alias for /ipfs

whyrusleeping:
In order to finally upgrade from using /ipfs everywhere to using /p2p as we should be, we need to make /p2p be an alias of /ipfs so that the upgrade path can happen easily.

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -12,8 +12,8 @@ code,	size,	name,	comment
 301,	0,	udt,
 302,	0,	utp,
 400,	V,	unix,
-420,	V,	p2p,	preferred over /ipfs
-421,	V,	ipfs,	equal to /p2p
+421,	V,	p2p,	preferred over /ipfs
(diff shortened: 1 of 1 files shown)


whyrusleeping:
@diasdavid says "just merge it, quick"

Arachnid:
Is  the intention of the multiprotocol/ipfs team that `p2p` be used for all content addressed systems, or should other systems such as swarm apply for their own protocol identifiers?

whyrusleeping:
@Arachnid Multiaddrs are used to reference specific processes, as opposed to content. The /p2p/ specifies that the thing we're trying to dial with the multiaddr speaks libp2p, and the contents of the protocol (the QmFooBar bit) specify the peer ID of the peer we expect to find at the other end of that pipe.

Arachnid:
So multiaddr doesn't support content-addressing? What's the argument to a `p2p` or `ipfs` type multiaddr? It appears to expect a hash.

PR #68, opened 2018-06-16
protocols.csv: Add "ip6%" multiaddr (IPv6 with zone ID)

mwnx:
This is a generalization of the "ip6" multiaddr which accepts an
optional zone ID as specified in [rfc4007].

[rfc4007] https://tools.ietf.org/html/rfc4007

---

One probably controversial point is that I'm using a non alphanumeric character ('%') in the protocol string. I'll be happy to change it to a 'z' if there is the slightest objection.

See also:
- https://github.com/mwnx/go-multiaddr/tree/ip6z
- https://github.com/mwnx/go-multiaddr-net/tree/ip6z
- https://github.com/mwnx/multicodec/tree/ip6z

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -1,10 +1,11 @@
-code,	size,	name,	comment
+code,	size,	name,		comment
 4,	32,	ip4,
 6,	16,	tcp,
 17,	16,	udp,
 33,	16,	dccp,
 41,	128,	ip6,
-53,	V,	dns,	reserved
+42,	V,	ip6zone,	rfc4007 IPv6 zone
+53,	V,	dns,		reserved
 54,	V,	dns4,
 55,	V,	dns6,
 56,	V,	dnsaddr,
@@ -12,8 +13,8 @@ code,	size,	name,	comment
 301,	0,	udt,
 302,	0,	utp,
 400,	V,	unix,
-421,	V,	p2p,	preferred over /ipfs
-421,	V,	ipfs,	backwards compatibility; equivalent to /p2p
+421,	V,	p2p,		preferred over /ipfs
+421,	V,	ipfs,		backwards compatibility; equivalent to /p2p
 444,	96,	onion,
 460,	0,	quic,
 480,	0,	http,


mwnx:
Another point I'm not sure about: I couldn't figure out any pattern or logic to the protocol code assignment so I just went with IP6 + 1.

Stebalien:
Yeah, I'd rather have `ip6z` or even `ip6ll` (link local). However, there's an alternative we should consider:

```
/link/lo0/ip6/.../tcp/1234
```

This would mean "connect to link X, and from there, connect to IP Y, ...". Thoughts?

/cc @lgierth you generally have opinions on multiaddr stuff.

lgierth:
Why not make /ip6 accept the zone-id? It's a valid part of an IPv6 address after all. (Just making sure I'm not missing anything while going through my backlog.)

Stebalien:
> Why not make /ip6 accept the zone-id? It's a valid part of an IPv6 address after all. (Just making sure I'm not missing anything while going through my backlog.)

The argument length is currently fixed so that would break existing addresses.

> It's a valid part of an IPv6 address after all.

It's a part of a valid textual IPv6 address but there's no place for it in the binary representation.

lgierth:
Got it -- that makes sense. It's also in line with `/ipcidr` being separate.

Name bikeshed: /ipzone? The additional characters don't seem too scary, especially given they're not sent over the network anyway.

mwnx:
I had considered adding a separate zoneid "multiaddr" code, but there were a few points I didn't quite like about it:

1. The zone ID is not a protocol. Not sure that matters in the end...

2. Where would the zone ID even go? Before or after the ip6 entry? Before seems reasonable since the zone ID represents an interface, which I guess is lower down in the IP stack...

3. The `zoneid` multiaddr protocol could only be combined with ip6 and nothing else. @Stebalien You seem to be suggesting adding a more general `link` "protocol" (that might for instance further restrict the interface to use in some circumstances I guess) but I feel like that would constitute an overreach of multiaddr's role: it would require multiaddr implementations to add some logic to distinguish between the ipv6 zone ID case and other cases and perform different network library/system calls depending on the case, and I'm pretty sure the exact semantics of both cases would end up differing.

4. Interaction and consistency with other non-ipfs/libp2p-related systems and tools. Many other interfaces (APIs, tools) consider an ipv6 address to be an ipv6 address + an optional zone ID. This includes go's `net` library and certain unix/posix functions I believe.

I'll admit these arguments are all quite weak, so if there is otherwise a consensus against the proposal as currently implemented, I'll make the necessary changes. As already stated in point 3., the only thing I strongly disagree with is the addition of a more general `link` "protocol". I would rather have an ipv6-specific `zoneid` "protocol" (or `zone`?, or `ip6zone` to avoid future collisions/confusion?).

mwnx:
So, what do you think? Would you accept a patch adding an `ip6zone` protocol?

Stebalien:
> The zone ID is not a protocol. Not sure that matters in the end...

True, but we don't using `dns{4,6,addr}` as "protocols" either. A multiaddr is basically just a set of steps describing how to build up a connection. In this case, "zone" would just mean "scope any subsequent IP addresses to the given zone".

> Where would the zone ID even go? Before or after the ip6 entry? Before seems reasonable since the zone ID represents an interface, which I guess is lower down in the IP stack...

Yes, before. It'd be `/link/lo/ip6/::1/tcp/8080`.

> it would require multiaddr implementations to add some logic to distinguish between the ipv6 zone ID case and other cases and perform different network library/system calls depending on the case, and I'm pretty sure the exact semantics of both cases would end up differing.

So, I'm pretty sure that the zone ID always corresponds to an interface number (on all platforms). `link` may not be the correct term (maybe `iface`?) but I'd like to generalize this as much as possible.

> Interaction and consistency with other non-ipfs/libp2p-related systems and tools. Many other interfaces (APIs, tools) consider an ipv6 address to be an ipv6 address + an optional zone ID. This includes go's net library and certain unix/posix functions I believe.

So... this quite often isn't the case. For example, in go, the IP struct doesn't have a place for the Zone. Instead, you have to use the IPAddr struct. This really is a general problem because the binary representation of an IPv6 address doesn't encode the zone.

---

FYI, an alternative is to just do what BSD does and store the zone ID in second 16bit word of the address. This will never conflict with anything as, in "proper" IPv6, this word must be 0. Actually, this may be the way to go. 

I'm not suggesting we *display* it there or put it there in the string format. Basically, we'd take an address in the form `/ip6/fe80::...%0/...` and serialize it to binary as `fe80:0:...`. Old versions may display these addresses as `/ip6/fe80:x:/...`, but that's not really a problem.

Thoughts?

Stebalien:
Basically, I'd like to avoid having *two* ipv6 protocols.

mwnx:
> FYI, an alternative is to just do what BSD does and store the zone
> ID in second 16bit word of the address. This will never conflict
> with anything as, in "proper" IPv6, this word must be 0. Actually,
> this may be the way to go. 
> 
> I'm not suggesting we *display* it there or put it there in the
> string format. Basically, we'd take an address in the form
> `/ip6/fe80::...%0/...` and serialize it to binary as `fe80:0:...`.
> Old versions may display these addresses as `/ip6/fe80:x:/...`,
> but that's not really a problem.
> 
> Thoughts?

I was aware of this BSD trick, however, I can see three major
reasons against using it in multiaddr:

1. I don't believe the RFC (rfc4007) specifies a maximum length for
   the numeric zone ID. There is in principle nothing preventing an
   operating system from creating interfaces with arbitrarily large
   IDs, such as `(uint64_t)-1`, although the RFC does refer to these
   as indexes, which would _suggest_ that they should be allocated
   sequentially. Better not to take any risks I'd say.

2. More fundamentally, I can already imagine two situations where
   the user wishes to specify the zone by name rather than by index
   and where the translation from name to index must be delayed:

   a. The user is a remote client sending commands to a server in
      non-textual form. These commands might include binary-encoded
      multiaddresses, which may include IPv6 addresses with zones.
      There would be no way for the client's multiaddr library to
      know how to translate the names to indexes.

   b. The interface might not yet be available. This could very
      easily happen in a processes that launches containers for
      instance.

3. I haven't read all the relevant RFCs but as far as I can tell,
   there's no guarantee that only `fe80::/64` link-local addresses
   can have a (non-default) zone. Quite the opposite, rfc4007 talks
   quite a bit about multicast addresses.

In conclusion, I will be implementing a `zone` "protocol" (probably
this week).

Do you think it would make sense (fit in with multiaddr's design
principles) to add some syntactic sugar translating
`/ip6/<ip6>%<zone>` to `/zone/<zone>/ip6/<ip6>`? I feel like it
would make for a better user experience in cases where human input
is expected, but is it worth the added complexity? What about the
reverse translation?

-- 
mwnx
GPG: AEC9 554B 07BD F60D 75A3  AF6A 44E8 E4D4 0312 C726

Stebalien:
> I was aware of this BSD trick, however, I can see three major reasons against using it in multiaddr:

All very good points.

> Do you think it would make sense (fit in with multiaddr's design principles) to add some syntactic sugar translating `/ip6/<ip6>%<zone>` to `/zone/<zone>/ip6/<ip6>`? I feel like it would make for a better user experience in cases where human input is expected, but is it worth the added complexity? What about the reverse translation?

If we go with the `/zone/` prefix, we wouldn't want to allow the `%` syntax.

(although I'd still call it something like `iface` as, e.g., `lo` is an interface, not a zone)

Stebalien:
@diasdavid @whyrusleeping @lgierth 

The proposal here is to add a new protocol, `/zone` (or `/iface` or `/link`) that allows one to specify the IPv6 "zone". This would be used as `/iface/eth0/ip6/ff80::3/tcp/1234`.

The alternatives were:

1. Introduce a *replacement* protocol for IPv6 that supports the `%zone` suffix. My objection was that this would introduce two IPv6 protocols.
2. Use the BSD and friends hack of sticking the zone in an unused 16bit word inside the encoded IPv6 address. However, as pointed out by @mwnx,
    1. We'd have to resolve interface names to numbers up-front when converting to a string to the binary representation. IMO, we may want to do this anyways and just make it a part of the string to binary conversion but that's something we can discuss after we've figured this step out.
    2. We'd only be able to support 16bit zone IDs. Given that the BSDs (including MacOS) do it this way, I'm not sure how much of an issue this would be *however*, it might mean we'd have to introduce a new syntax later *anyways* to deal with larger zone IDs.
    3. This would only support link-local addresses. IMO, this is the main reason not to do it this way.

Given that this is a decision that'll impact us down the road, you should take a look and think about it.

mwnx:
Well, I finally got to it (one week late, sorry). I've force-pushed on all four projects. I went with the "zone" prefix solution. And yes, I do think it should be called "zone" since that is the vocabulary used in the IETF standard (rfc4007).

Stebalien:
> And yes, I do think it should be called "zone" since that is the vocabulary used in the IETF standard (rfc4007).

Hm. I see. There's actually a distinction between link-local and interface-local so either interface or link would definitely be wrong.

mwnx:
So, uh, any chance of this being merged?

Stebalien:
I was hoping to get a signoff from either @lgierth or @diasdavid.

Stebalien:
Note: It looks like the multicodec definition still needs to be updated.

mwnx:
On Sun, Jul 29, 2018 at 03:10:56PM -0700, Steven Allen wrote:
> Note: It looks like the multicodec definition still needs to be updated.

Indeed. I updated the commit message but forgot to update the file.
Done.

-- 
mwnx
GPG: AEC9 554B 07BD F60D 75A3  AF6A 44E8 E4D4 0312 C726

Stebalien:
@mwnx need to push :smile:.

mwnx:
On Mon, Jul 30, 2018 at 08:14:49PM +0000, Steven Allen wrote:
> @mwnx need to push :smile:.

I pushed but I'd never made the pull request...

https://github.com/multiformats/multicodec/pull/79

-- 
mwnx
GPG: AEC9 554B 07BD F60D 75A3  AF6A 44E8 E4D4 0312 C726

Stebalien:
This PR also needs to be updated.

mwnx:
On Mon, Jul 30, 2018 at 10:43:55PM +0000, Steven Allen wrote:
> This PR also needs to be updated.

Oops again. And... fixed.

-- 
mwnx
GPG: AEC9 554B 07BD F60D 75A3  AF6A 44E8 E4D4 0312 C726

mwnx:
Bump :)

Stebalien:
Still trying to get at least one other PL member to sign off on this.

protocols.csv
```diff
@@ -4,6 +4,7 @@ code,	size,	name,	comment
 17,	16,	udp,
 33,	16,	dccp,
 41,	128,	ip6,
+42,	V,	zone,	rfc4007 IPv6 zone
```
diasdavid: Make it more explicit and use `ip6zone` for the name instead

diasdavid:
Pinging the Lead Maintainer of the `multiaddr` JS module, @VictorBjelkholm and js-libp2p devs, @jacobheun & @vasco-santos to review this proposal.

jacobheun:
This looks good to me and the change to `ip6zone` makes sense.

protocols.csv
lgierth: Yes let's go with `/ip6zone` -- these zones are about IPv6 only at the moment, and at the moment I don't see us generalizing this enough to fit other cases (similar to /dns4,6 vs. /dns).

vasco-santos:
Makes sense to me! 👍

mwnx:
On Wed, Aug 22, 2018 at 12:46:57AM -0700, David Dias wrote:
> Make it more explicit and use `ip6zone` for the name instead

Done. Force-pushed on all 4 repositories.

-- 
mwnx
GPG: AEC9 554B 07BD F60D 75A3  AF6A 44E8 E4D4 0312 C726

Stebalien:
I guess we can still say "faster than the IETF", but that's not saying much.

Stebalien:
@mwnx mind making PRs against go-multiaddr now?

mwnx:
On Mon, Sep 10, 2018 at 04:31:50PM -0700, Steven Allen wrote:
> @mwnx mind making PRs against go-multiaddr now?

Just fixed the conflict in my go-multiaddr PR [1]. See also the
go-multiaddr-net PR [2].

[1] https://github.com/multiformats/go-multiaddr/pull/69
[2] https://github.com/multiformats/go-multiaddr-net/pull/42

-- 
mwnx
GPG: AEC9 554B 07BD F60D 75A3  AF6A 44E8 E4D4 0312 C726

PR #71, opened 2018-07-21
RFC: Add a memory protocol

tomaka:
In rust-libp2p, for unit testing purposes, we've created an implementation of `Transport` that lets you create multiple objects that can connect to each other.
With this, you can create a client and a server and test whether they work fine together, without having to open a TCP connection on localhost.

But the question was: which multiaddr should be passed when we "dial" or "listen" on this transport, for which we already known it will be received by the other party?

This PR adds the possibility to have a multiaddr `/memory` (with no parameter), which means that the dialer "magically" knows how to reach the listener.
On the theoretical level, a multiaddress like `/memory/0x14b8a90c` makes sense in my opinion. In practice that would be obviously too unsafe, so we opted for `/memory` alone.

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -24,3 +24,4 @@ code,	size,	name,	comment
 275,	0,	p2p-webrtc-star,
 276,	0,	p2p-webrtc-direct,
 290,	0,	p2p-circuit,
+777,	0,	memory,


alexander255:
Would this not end being a transport using `pipe(2)`? If so, would an address of the form `/fd/<fd_in>/<fd_out>` make more sense?

Maybe I don't understand what you mean by “magically received”. How do these transports exchange data then? If you create them without any kind of dialing, why do you need a multiaddr at all? An example would be appreciated I guess. :slightly_smiling_face:

lgierth:
This is a transport for test environments, where the endpoints share a memory space, and thus don't need to do any actual networking or file descriptors.

Gonna merge as I see it's in rust-libp2p: https://github.com/libp2p/rust-libp2p/blob/master/misc/multiaddr/src/protocol.rs

lgierth:
Actually, @tomaka could you claim the code in multiformats/multicodec/table.csv first?

bigs:
furthermore, this transport is useful when a peer might dial itself. we're working on this in libp2p/go-libp2p#638. this seems good to me

bigs:
@Stebalien how do we feel about this? i'm ready to merge as-is.

@tomaka the only extension i've thought about is adding a string component representing a "slot" name which would be like a virtual port. this avoids the safety issues of pointers, but ultimately i can't imagine why you'd need more than one transport. especially when multistream's entire purpose is essentially to provide such slots

tomaka:
> @tomaka the only extension i've thought about is adding a string component representing a "slot" name which would be like a virtual port. this avoids the safety issues of pointers, but ultimately i can't imagine why you'd need more than one transport. especially when multistream's entire purpose is essentially to provide such slots

Since I opened this PR (10 months ago), we added a 64bits integer as parameter.

In the code, our memory transport is now global (just like TCP ports are global, for instance). Tests generate a random "memory port number", and we assume that 64bits is large enough to not have collisions.

lgierth:
Let's just make the value an arbitrary one. There's no interop or compat concerns since this stuff always stays in the same process.

raulk:
Thoughts:

* We do want the ability to multiplex on memory -- just like a TCP listener can listen on many ports, so should a memory transport.
* Let's use an arbitrary value to address the memory "bridge" -- the mechanism is implementation-specific and we can't assume that we'll have access to a concrete memory address. Example: languages that abstract memory like Java, Python.
* We should ensure that memory addresses don't escape the local host – urgh.

raulk:
Isn't this table deprecated in favour of multiformats/multicodec?

lgierth:
> Isn't this table deprecated in favour of multiformats/multicodec?

Yes, should be removed, will be taken care of soon

bigs:
> Thoughts:
> 
>     * We do want the ability to multiplex on memory -- just like a TCP listener can listen on many ports, so should a memory transport.

as stands now, i just open a new golang "pipe" (backed by channels) per "stream" so it has stream support without any overhead of security or a muxer.

>     * Let's use an arbitrary, opaque value to address the memory "bridge" -- the mechanism is implementation-specific and we can't assume that we'll have access to a concrete memory address. Example: languages that abstract memory like Java, Python.

i don't fully understand why we need "addresses". you still have to dial by peer ID. i think we could have one memory transport per peer id and be fine with it. i don't see what we gain from being able to have slots after thinking about it more.

>     * We should ensure that memory addresses don't escape the local host – urgh.

this is circumvented by using either string slots, arbitrary int slots (like @tomaka mentioned) or no slots at all (my preference.)

Stebalien:
> Yes, should be removed, will be taken care of soon

The issue is that the multicodec table can't record argument lengths. We'd need to make that table extensible (json?).

tomaka:
> Let's use an arbitrary, opaque value to address the memory "bridge" -- the mechanism is implementation-specific and we can't assume that we'll have access to a concrete memory address. Example: languages that abstract memory like Java, Python.

Exposing the memory address (and most importantly, interpreting it as a pointer) would be an extremely bad idea in my opinion.
The parameter, if any, has to be opaque.

tomaka:
In rust-libp2p, we're using memory addresses for testing, not for self-dialing.
For example we simulate 3 nodes A, B, C, connect both B and C to A, then check whether B learns about C and vice-versa. This is why we need some virtual port number, and not just a `/memory` address that assumes that the destination is implicitly known.

bigs:
@tomaka makes sense. i'm happy with uint64 or a string, and since you're already using uint64 there's less friction to keep going that route.

thoughts @raulk @Stebalien ?

tomaka:
> since you're already using uint64 there's less friction to keep going that route.

Since memory addresses should never leave the machine, it's not critical if Go uses for example a string instead.

raulk:
Memory addresses should be truly opaque, in the sense that we don’t care if their value is uint64 or a byte string. It’s up to the implementation to interpret that value. Interoperability is not required because these addresses must never escape the local stack.

bigs:
merged. i may update it a bit (it was in a fork) to specify either 64bit int argument.

lgierth:
Let's make the value length-prefixed - we should really stop adding protocols with a fixed length. It's out-of-band information (!) and the root cause for why we can't deal well with new or custom protocols.

raulk:
@bigs – length-prefixing (as @lgierth says) is the appropriate way to represent _opaque_ values. As we discussed above, implementations are free to choose the type and length of their memory references. We shouldn't enforce `(u)int64`.

bigs:
ah sorry here, these didn't make it. i'll patch in my branch.

PR #74, opened 2018-11-22
Add bluetooth protocol

tomaka:
The parameter is 6 bytes indicating the bluetooth device ID to connect to.

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -25,3 +25,4 @@ code,	size,	name,		comment
 275,	0,	p2p-webrtc-star,
 276,	0,	p2p-webrtc-direct,
 290,	0,	p2p-circuit,
+802,	48,	bluetooth


alexander255:
@lanzafame: Yes, a Bluetooth device ID is a MAC address – and so is the MAC address used in Ethernet. How would then know which one is meant if I tell you `/mac/22:44:33:66:FF:11`? You wouldn't, because the MAC address is just the addressing format and not the protocol. You still would have no idea what protocol to use.
Since you suggested aliasing `bluetooth` to `mac`: Does your suggesting also include aliasing `ether` to `mac`? Because that, aside from causing perfect confusion, would be perfectly in line with what you have suggested above.

tomaka:
Using `mac` instead of `bluetooth` would be similar to using `u16` instead of `tcp` and `udp`.

alexander255:
@tomaka: You nailed it! :grin:

lanzafame:
@tomaka @alexander255 fair and valid points :smile:

alexander255:
@lanzafame: I thought about this again however and while the point about `mac` stands, I realized why adding `/bluetooth` is feels like a bad idea (maybe that's what upset you as well leading to the less-than-optimal suggestion of yours): Bluetooth isn't a single protocol, but rather a complete protocol suite + application-level descriptions of expected endpoint behaviour and UX (aka “profiles”).

After some digging in the Bluetooth specification however, I've come to realize that the Bluetooth/MAC addresses are part of the very core of the protocol specification. Nothing will work without them and the parts that use them don't even have a separate name and the relevant sections in the spec are simply titled “Link Controller operation” (cf Vol 2 / Part B / Section 8) and “Physical links” (cf Vol 2 / Part B / Section 3). As such calling the protocol `/bluetooth` isn't wrong, but maybe `/bluetooth-link` or `/bluetooth-phys` would be more descriptive in that that this protocol is only about the point-to-point aspect of the protocol suite?

Either way, this won't give you an application-level connection that you could use as such. For this, other higher-level protocols such as L2CAP (providing UDP-like multiplexing) and RFCOMM (providing a TCP-like reliable transport on top of L2CAP) are needed.

tomaka:
Should we add `/l2cap/<port>` and `/rfcomm/<port>` as well?

alexander255:
@tomaka: Yes, exactly. Note however that establishing an RFComm connection would then look something like this:
`/bluetooth-phys/<mac>/l2cap/<channel>/rfcomm/<srv-channel>`, where the L2CAP channel number is almost always *3* (the reserved value for RFComm) when using RFComm.

This does not address how one *finds* these values however, since that's what SDP (Bluetooth DNS :slightly_smiling_face:) is used for. So, to describe service resolution we'd need *that* as well. For SPP (the Bluetooth profile describing emulated serial ports) that would probably be along the lines of:
`/bluetooth-phys/<mac>/sdp/SerialPort` or
`/bluetooth-phys/<mac>/sdp/0x1101`
I may be wrong on that, so don't take my word for it. Anyways, I'm done reading Bluetooth specs for today.

(In case you're wondering, the numbers are from [this page](https://www.bluetooth.com/specifications/assigned-numbers/service-discovery).)

tomaka:
To me SDP should be part of the discovery mechanism. It's not a reason not to add it to multiaddr, but I don't think it's needed.

Anyway, I've opened this PR because I needed "/bluetooth" addresses in a local prototype, and therefore had to assign a number. If we want to do things correctly, let's add everything at once later.

tomaka:
I just hope that there's not someone else writing a prototype and assigning different numbers for Bluetooth protocols.

PR #76, opened 2018-11-26
Add dtls multiaddr

### LINKED ISSUE #75: Add DTLS protocol
We're considering building DTLS transports in libp2p/libp2p#49. Therefore, I suggest to add a 'dtls' multiaddr protocol. I'd like to have some input on the following:
1. We don't really need a 'value' part of the multiaddr to establish a connection. However, we could add a certificate fingerprint as a form of authentication. Is this a good use of the 'value'? Can this be made optional?
1. Should I open a PR on this, the multicodec or both repo's?

backkem:
This PR adds a dtls multiaddr to resolve #75. I've not foreseen a value part for now. ``0x20`` also seems free in ``multiformats/multicodec``. Let me know if I should also open a PR on that repo.

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -2,6 +2,7 @@ code,	size,	name,		comment
 4,	32,	ip4,
 6,	16,	tcp,
 17,	16,	udp,
+32,	0,	dtls,
 33,	16,	dccp,
 41,	128,	ip6,
 42,	V,	ip6zone,	rfc4007 IPv6 zone


Stebalien:
So, regardless of whether or not go-libp2p can *currently* use dtls, we can still add the protocol code. However,

1. We should first PR against https://github.com/multiformats/multicodec as that's the *complete* list of codecs.
2. We should probably pick a two-byte multicodec as we have very few single-byte multicodecs available.

PR #77, opened 2018-11-27
Register new address types for Tor and i2p

eyedeekay:
I hope I am doing this in the right order. I have a branch of go-multiaddr over here https://github.com/RTradeLtd/go-multiaddr where I've been experimenting with onionv3 and i2p base64 addresses. They are working nicely and I have integrated them into the tests, now I believe I need to register the protocols before I can begin upstreaming them.

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -17,8 +17,8 @@ code,	size,	name,		comment
 421,	V,	ipfs,		backwards compatibility; equivalent to /p2p
 444,	96,	onion,
 445,    296, onion3
-446,    276, garlict
-447,    276, garlicu
+446,    3112, garlict
+447,    3112, garlicu
 460,	0,	quic,
 480,	0,	http,
 443,	0,	https,
--- a/protocols.csv
+++ b/protocols.csv
@@ -17,8 +17,8 @@ code,	size,	name,		comment
 421,	V,	ipfs,		backwards compatibility; equivalent to /p2p
 444,	96,	onion,
 445,    296, onion3
-446,    3112, garlict
-447,    3112, garlicu
+446,    3096, garlic64
+447,    256, garlic32
 460,	0,	quic,
 480,	0,	http,
 443,	0,	https,
--- a/protocols.csv
+++ b/protocols.csv
@@ -16,6 +16,9 @@ code,	size,	name,		comment
 421,	V,	p2p,		preferred over /ipfs
 421,	V,	ipfs,		backwards compatibility; equivalent to /p2p
 444,	96,	onion,
+445,    296, onion3
+446,    276, garlict
+447,    276, garlicu
 460,	0,	quic,
 480,	0,	http,
 443,	0,	https,
--- a/protocols.csv
+++ b/protocols.csv
@@ -17,7 +17,7 @@ code,	size,	name,		comment
 421,	V,	ipfs,		backwards compatibility; equivalent to /p2p
 444,	96,	onion,
 445,    296, onion3
-446,    3096, garlic64
+446,    V, garlic64
 460,	0,	quic,
 480,	0,	http,
 443,	0,	https,
--- a/protocols.csv
+++ b/protocols.csv
@@ -18,7 +18,6 @@ code,	size,	name,		comment
 444,	96,	onion,
 445,    296, onion3
 446,    3096, garlic64
-447,    256, garlic32
 460,	0,	quic,
 480,	0,	http,
 443,	0,	https,


Stebalien:
Instead of garlicu and garlict, maybe just `/garlic/.../tcp/...`? I know it doesn't really mirror onion but, well, in retrospect onion probably should have used that syntax as well. IMO, it definitely makes sense *here* because we support both.

Also, what about base32 addresses? I'm wondering if we should try supporting both in the same protocol or or define multiple protocols.

eyedeekay:
Thanks for the feedback, I can switch it to the garlic/tcp format instead. If you would like, I can also do it for the onionv3 addresses too.

I chose to use base64 addresses simply because it's easier to go from base64 to base32 in an application, especially with sam3, which is the existing go library for talking to i2p. For this reason so far I've been giving applications a base64 and letting the application decide how to handle it. My issue with putting them in the same protocol is how you access both the base64 and base32 in the same protocol without having to give whichever multiaddr implementation i2p-specific functionality to get the base32 from the base64 and make it available. So from that perspective, I would lean toward having a separate protocol for base32 addresses.

Stebalien:
> If you would like, I can also do it for the onionv3 addresses too.

I'd leave *that* as-is. It should probably mirror the current onion transport.

> base32...

Makes sense to me.

eyedeekay:
I changed it from /garlict and /garlicu to  /garlic64 and /garlic32 and removed the port from the length. The changes are also reflected in my submission to go-multiaddr now.

Stebalien:
> garlic32

Do you actually *need* garlic32 addresses right now?

I'm a *bit* worried that those names will lead users to think that it's OK to take a garlic32 address, re-encode it using base64, and then use it in a garlic64 address. However, the base encoding is just a surface difference. The real difference is that the *former* is the hash of the key (and needs to be resolved) and the *latter* is the key itself.

So, I'm kind of tempted to just punt. I'm not sure if we should have *three* protocols (`/garlic`, one for `.i2p` sites (`/i2p`?), one for b32 addresses) or a garlic one and a single i2p one that supports *both* `.i2p` sites *and* b32 addresses.

Basically, this distinction is more like a dns4 versus ip4 than anything else.

eyedeekay:
I only need base64 and am happy to defer adding base32 until such time as they are needed. I honestly hadn't given as much thought to the human-readable links yet, but I was working from the assumption that if someone needed them, they would require a third protocol. I will reconsider that in the meantime.

Stebalien:
Looking through the i2p documentation a bit, are you sure garlic64 addresses are *always* 3096 bits long? According to https://geti2p.net/en/docs/naming#lookup:

> All destinations in I2P are 516-byte (or longer) keys. (To be more precise, it is a 256-byte public key plus a 128-byte signing key plus a 3-or-more byte certificate, which in Base64 representation is 516 or more bytes. Non-null Certificates are in use now for signature type indication. Therefore, certificates in recently-generated destinations are more than 3 bytes.

I believe we'll need to make this a variable-length address (V).

---

Aside: My last post wandered a bit. I meant to suggest that we go back to just `/garlic` (no 64) for now as I find talking about *entirely different* address types in terms of the base in which we happen to be encoding them really confusing. However, it looks like the official i2p documentation calls these addresses "base64" and "base32" so I guess it's fine to leave it as-is. I'll leave it up to you as I'm not an expert on this subject.

eyedeekay:
In practice I've only seen the 516(and it would appear that this is all sam3 can give me right now), but the documentation you linked says that it's supposed to be able to handle from 516-616, so I'll make it variable-length.

My thought on the /garlic vs /garlic64 decision is mostly related to the mechanics of the human-readable addressbook, which is not DNS-like and in fact it's not at all clear how applications other than the i2p HTTP proxy are supposed to communicate with it. Applications that use SAM can use pre-existing human-readable addresses in the addressbook, but have no way of looking up new ones unless they choose to manage their own. Adding new addresses is very tied to the use of the HTTP proxy to browse eepSites(Which is arguably it's intended single use), where clicking on an addresshelper link will pop up a screen which asks the user directly "Would you like to add X address to your local addressbook." One can programmatically query a jump service(Upstream provider of human-readable addresses) and add it to a list in one's application, but it won't be in the included addressbook application where it would become available to other apps, because silently adding addresses to the shared human-readable addressbook in the background would be a pretty dangerous proposition. Base32 addresses obviously don't have all these complications. I think that there are remarkable differences in the properties of base32 and human-readable addresses, including how you get them and how you manage them, which may merit considering them separate things. So /garlic64 /garlic32 and something else, maybe /garlicjump or just /garlic?

Stebalien:
> My thought on the /garlic vs /garlic64

SGTM. Let's add garlic64 for now.

PR #78, opened 2018-11-28
fix udp code

Stebalien:
This was changed ages ago but not updated here. The current code *here*
conflicts with SHA1 (in the multicodec table).

Ideally, we wouldn't duplicate these tables but *this* table lists *sizes*. We
could also just remove the numbers from this table and *only* list names/sizes
but we can do that later.


Stebalien:
See: https://github.com/multiformats/multicodec/pull/16

PR #82, opened 2018-12-14
Add comma to properly format CSV

EvanHahn:
The `onion3` row didn't end with a comma so the CSV wasn't technically formatted properly. This should fix that.


Stebalien:
Thanks for catching that.

PR #90, opened 2019-04-05
add garlic32 addresses

eyedeekay:
I now need to use multiaddr to keep track of i2p base32 addresses, so I've added them to the go library to be compatible with the current b32.i2p specification and to handle the near-future base32 for encrypted leasesets specification. I just used the next sequential number after garlic64.


Stebalien:
Could you also make a PR against https://github.com/multiformats/multicodec?

eyedeekay:
Yup: https://github.com/multiformats/multicodec/pull/126

PR #91, opened 2019-05-31
Update definition of memory transport

bigs:


lgierth:
See https://github.com/multiformats/multiaddr/pull/71#issuecomment-497748677 -- the value should be length-prefixed instead of fixed-length.

PR #92, opened 2019-05-31
Make memory transport length-prefixed

bigs:
Apologies for the sloppiness, eager to get this going. I think for the golang implementation these bytes will be interpreted as a uint64.


tomaka:
I just want to point out that if the payload contains a `/`, then you have an ambiguity in the text format. This is why I went with numbers only in https://github.com/multiformats/multiaddr/pull/71

PR #97, opened 2019-07-23
Add ZeroNet site address

filips123:
Add [ZeroNet](https://zeronet.io/) site address to protocol codes. ZeroNet uses old-style Bitcoin addresses as site addresses.

Also see multiformats/multicodec#140.

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -30,3 +30,4 @@ code,	size,	name,	comment
 276,	0,	p2p-webrtc-direct,
 290,	0,	p2p-circuit,
 777,	V, memory, in memory transport for self-dialing and testing; arbitrary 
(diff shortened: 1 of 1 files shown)


alexander255:
I vote to reject this as part of `multiaddr` as Zeronet addresses are used as a means to address *content*, rather then *endpoints* (devices/servers/services/…).

Adding this to https://github.com/multiformats/multicodec seems reasonable however. :+1:

filips123:
@alexander255 I agree with you. Which type tag should I use instead in multiformats/multicodec#140?

Stebalien:
Let's move the conversation over to https://github.com/multiformats/multicodec/pull/140. We can reopen this PR if necessary.

Issue #94, opened 2019-07-09
DNS name binary encoding


alexander255:
As mentioned in #22, `/dns`, `/dns4`, `/dns6` & `/dnsaddr` do not currently have an official binary encoding defined.

In [`py-multiaddr`](https://github.com/multiformats/py-multiaddr/blob/master/multiaddr/protocols.py#L130-L133) we currently use Unicode for the text representation and IDNA-2008/Punnycode for the binary representation. Obviously this distinction is only relevant for domains with labels containing non-ASCII characters.

How is this handled in other implementations? Is the current behaviour something that could be standardized or is another behaviour more desirable?

Stebalien:
I believe the current de-facto standard in go and js is to just encode the unicode to utf-8. IIRC, punnycode is only _really_ needed for compatibility with the DNS protocol.

However, this may have canonization issues.

alexander255:
> However, this may have canonization issues.

Canonicalizing domain name strings is specified by [UTS46](https://www.unicode.org/reports/tr46/#Processing). In particular that specification requires one to [ship a mapping table] and decode all domain name labels starting with `xn--`using [Punnycode](http://ietf.org/rfc/rfc3492.txt) to validate & lower-case their contents. So while could store the end result of this process as UTF-8 rather then IDNA, we'd still have to implement/ship most IDNA stuff anyways to ensure canonicalization of the generated values.

Stebalien:
(belatedly)

I believe most DNS resolvers will do this internally. I'm fine saying that any domain in a `/dns` string must have already been converted from punycode to utf-8.

(unless I'm missing something important)

alexander255:
@Stebalien and me went for standardizing UTF-8 encoding with UTS-46 normalization/canonicalization in #101.

Issue #70, opened 2018-07-18
Handle unparsable multiaddrs


Stebalien:
Due to the fact that binary multiaddrs don't include protocol definitions, we can't parse them unless we know all the relevant protocols.

However, we *can* parse a prefix. Therefore, I'd like to propose a special, string-only "unknown" protocol that takes a single multibase encoded argument. That is: `/ip4/1.2.3.4/tcp/123/unknown/bxyz". This would only exist the string format and would allow us to keep and use multiaddrs we don't fully understand.

This should fix some of the problems described in #6.

alanshaw:
A few things I'd like to explore:

1. How can we retain the original string? In terms of printing these addrs out, it's more useful to be able to print the representation that we recieved, not a `/unknown/whatever` even if it cannot be used. The use case is an old js-ipfs-api requesting `swarm.addrs` to be printed out on a webpage. It might not understand how to parse QUIC addrs but it's preferable to print the QUIC addr than an "unknown" addr
1. How can we distinguish between a multiaddr that is valid or not? If we can pass it any string protocol - how do we determine if it's valid (at least in our version of the codec table)? What would valid mean now? What do we do when passing a string/buffer to the multiaddr constructor?
1. If we create a "string only" multiaddr - what happens when we request the buffer version - throw error?

alanshaw:
@jacobheun it might be useful to have your opinion on this?

olizilla:
The requirement that a multiaddr can be represented as a buffer is tricky. Every user of multiaddr has a stale copy of the buffer-to-protocol mapping. It's trivial to syntacitically validate a multiaddr string, but hard to semantically validate that protocols are valid if the definition of valid is their existance in the mapping table. 

We will keep running into situations where peers and apps and api clients are running with different versions of multiaddr, with different mapping tables, and older ones unable to parse newer ones. 

Having a string only format for multiaddr that doesn't guarantee it can be converted to a buffer would at least tick the points 1 to 4 on the problem statement https://github.com/multiformats/multiaddr#introduction

Stebalien:
I think we're talking past each other a bit here. My primary concern is relay dialing (which should have been in the issue description...). That is, I need to be able to use a multiaddr prefix (encoded) without the rest. 

For example: `/ip4/1.2.3.4/tcp/1234/p2p/QmRelay/p2p-circuit/ip4/tcp/1234/special_protocol/p2p/QmTarget`.

[...]

PR #101, opened 2019-09-01
Document MultiAddr codecs implemented by py-multiaddr

alexander255:
Inspired by the Python code I wrote with lots of references to relevant specs.


alexander255:
@Stebalien @lgierth: Bump. What trick would I have to use to realistically get some discussion/feedback on this with everybody being so busy? :upside_down_face:

ShadowJonathan:
Bump, I believe this is a useful addition to the repo.

README.md
```diff
@@ -104,6 +104,166 @@ TODO: specify the encoding (byte-array to string) procedure
 
 TODO: specify the decoding (string to byte-array) procedure
 
+### Codecs
+
+Depending on the protocol of a Multiaddr item, different algorithms are used to
```
Stebalien:
```suggestion
Depending on the protocol of a Multiaddr component, different algorithms are used to
```

README.md
```diff
@@ -104,6 +104,166 @@ TODO: specify the encoding (byte-array to string) procedure
...
+
+In general empty values in the string representation are always disallowed unless
+explicitely noted otherwise. In case of conversion errors implementation must
+refuse to process the given string/binary value and report the error to the caller
+instead.
+
+While all codec types may be encoded using the standard variable length encoding,
+some may also allow being encoded into a specific static-length binary value
+without the extra length information if this is noted in the respective codec's
+description.
```
Stebalien: This isn't true in go (and probably not in other languages either). Only variable length codecs can use length prefixes.

README.md
```diff
@@ -104,6 +104,166 @@ TODO: specify the encoding (byte-array to string) procedure
...
+
+Encodes the given Unicode string using the system's local file system encoding.
+On Windows this encoding likely being UTF-16, while being UTF-8 on most other
+systems. It is up to the library to figure out the best encoding value for
+these kinds of strings.
+
+ * String → Binary: `str.encode(SYSTEM_FILESYSTEM_ENCODING)`
+ * Binary → String: `bytes.decode(SYSTEM_FILESYSTEM_ENCODING)`
+
+Protocols using the `fspath` encoding must not be shared between different hosts.
```
Stebalien: Go and JS use utf8 and forward slashes. It's up to the application to convert paths into the local operating system's preferred format.

README.md
```diff
@@ -104,6 +104,166 @@ TODO: specify the encoding (byte-array to string) procedure
...
+
+#### `idna`
```
Stebalien: https://github.com/multiformats/multiaddr/issues/94

README.md
```diff
@@ -104,6 +104,166 @@ TODO: specify the encoding (byte-array to string) procedure
...
+ * Binary → String:  
+   Convert the ASCII text string to Unicode according to the rules of [UTS-46 Section 4.3 (ToUnicode)](https://www.unicode.org/reports/tr46/#ToUnicode) using the same parameters as in step 1 of the *String → Binary* algorithm.
+
+Examples of libraries for performing the above steps include the [Python idna](https://pypi.org/project/idna/) library.
+
+#### `ip4`
+
+Encodes an IPv4 address according to the conventional [dot-decimal notation](https://en.wikipedia.org/wiki/Dot-decimal_notation) first specificed in [RFC 3986 section 3.2.2 page 20 § 2](https://tools.ietf.org/html/rfc3986#page-20).
+
+Protocols using this codec may encode it as binary value of exactly 4 bytes without
```
Stebalien:
```suggestion
Protocols using this codec must encode it as binary value of exactly 4 bytes without
```

README.md
```diff
@@ -104,6 +104,166 @@ TODO: specify the encoding (byte-array to string) procedure
...
+an extra length value.
+
+ * String → Binary:
```
Stebalien: Not sure if explicitly specifying an IPv4 address is necessary given the link to the spec.

README.md
```diff
@@ -104,6 +104,166 @@ TODO: specify the encoding (byte-array to string) procedure
...
+functions. Alternatively, using the BSD
+[`getaddrinfo`/`freeaddrinfo`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/getaddrinfo.html)
+and [`getnameinfo` with `NI_NUMERICHOST`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/getnameinfo.html)
+may be a viable alternative for some environments.
+
+### `onion`
+
+Encodes a [TOR rendezvous version 2 service pointer](https://gitweb.torproject.org/torspec.git/tree/rend-spec-v2.txt?id=471af27b55ff3894551109b45848f2ce1002441b#n525) (aka .onion-address) and exposed service port on that system.
+
+Protocols using this codec may encode it as binary value of exactly 12 bytes without
```
Stebalien:
```suggestion
Protocols using this codec must encode it as binary value of exactly 12 bytes without
```

README.md
```diff
@@ -104,6 +104,166 @@ TODO: specify the encoding (byte-array to string) procedure
...
+function and its equivalent bindings in many other languages. Similarily the POSIX
+[`inet_ntoa`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/inet_ntoa.html)
+function available in many languages implements the previously mentioned binary
+to string address transformation.
+
+#### `ip6`
+
+Encodes an IPv6 address according to the rules of [RFC 4291 section 2.2](https://tools.ietf.org/html/rfc4291#section-2.2) and [RFC 5962 section 4](https://tools.ietf.org/html/rfc5952#section-4).
+
+Protocols using this codec may encode it as binary value of exactly 16 bytes without
```
Stebalien:
```suggestion
Protocols using this codec must encode it as binary value of exactly 16 bytes without
```

README.md
```diff
@@ -104,6 +104,166 @@ TODO: specify the encoding (byte-array to string) procedure
...
+       `service_str = b32encode(service_bin)`
+    3. Convert the *port* part to text as specified by the [`uint16be`](#uint16be) codec.
+    4. Concatenate the result strings using a colon:  
+       `return service_str + ":" + port_str`
+
+### `p2p`
+
+Encodes a libp2p node address.
+
+TBD: Is this really always a base58btc encoded string of at least 5 characters in length!?
```
Stebalien: It's anything that parses as a peer ID (the canonical encoding is now actually a libp2p key _CID_): https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md.

README.md
```diff
@@ -104,6 +104,166 @@ TODO: specify the encoding (byte-array to string) procedure
...
+
+
+### `uint16be`
+
+Encodes an unsigned 16-bit integer value (such as a port number) in network byte
+order (big endian).
+
+Protocols using this codec may encode it as binary value of exactly 2 bytes without
```
Stebalien:
```suggestion
Protocols using this codec must encode it as binary value of exactly 2 bytes without
```

README.md
alexander255: At least the go implementation does not follow that spec at all but just expects everything to be base58btc: https://github.com/multiformats/go-multiaddr/blob/master/transcoders.go#L293-L315

If your saying that is just a bug to be fixed, I'll be happy to update this text and py-multiaddr accordingly. :slightly_smiling_face:

README.md
alexander255: It'd be a nice transitional step for making #70 possible for unknown binary formats. But I see that this is not the right place for discussing this. I'll move this to a different PR, so that we can merge this one…

README.md
alexander255: The linked spec is actually the URL spec that has an “aside” kind of note on how to *parse* IPv4 dotted-decimal addresses in an URL hostname. There is no formal spec on IPv4 textual address representation actually, so I thought reproducing the expected steps here might avoid ambiguity: For instance the POSIX `inet_addr` function would also allow `"127.0x000001"` as a valid input for `"127.0.0.1"`, so I wanted to be clear *which variations* are allowed by this spec and which are not.

README.md
ShadowJonathan: That spec also fails to mention that some IDs are still encoded with `1` as first character, which is also CIDv0

README.md
alexander255: Think about this again, what I'm actually trying to achieve with this is proper canonicalization of the input. And while IDNA encoding has all that there is no reason why can't use UTS-46 normalization with UTF-8 encoding. So how about this instead:

> Encodes the given Unicode representation to the UTF-8 character encoding ([RFC 3629 Section 3](https://tools.ietf.org/html/rfc3629#section-3)), while using the [UTS-46 / RFC 5890](https://tools.ietf.org/html/rfc5890) input normalization and processing rules for canonicalization.
> 
> * String → Binary:
>    1. If feasible, normalize and validate the given input string according to [UTS-46 Section 4 (Processing)](https://www.unicode.org/reports/tr46/#Processing) and [UTS-46 Section 4.1 (Validity Criteria)](https://www.unicode.org/reports/tr46/#Validity_Criteria) with the following parameters:
>       * UseSTD3ASCIIRules = true
>       * CheckHyphens = true
>       * CheckBidi = true
>       * CheckJoiners = true
>       * Transitional_Processing = false
>    2. Convert the Unicode string to the UTF-8 character encoding as per [RFC 3629 Section 3 §4](https://tools.ietf.org/html/rfc3629#section-3).
> * Binary → String:  
>   Convert the UTF-8 encoded binary string to Unicode according to the rules of [RFC 3629 Section 3 §6](https://tools.ietf.org/html/rfc3629#page-5).

Stebalien: That sounds like a great idea (and it fixes the bytewise comparison problem).

README.md
Stebalien: @alexander255 correct (well, not a bug so much as something we need to implement).

@ShadowJonathan also correct. I'll take a stab at fixing the spec.

alexander255: @Stebalien: I just tried to implement that spec in Python and there is no guidance in there what to do with binary CIDv0 IDs. I assume they are detected (similar to their string format) using `len(buf) == 34 and buf.startswith(b"\x12\x20")`?

ntninja:
@Stebalien: I've updated the description of `fspath` to leave the binary encoding completely unspecified. Can we merge this now?

PR #105, opened 2019-12-14
Implemented Elixir Multiaddr

aratz-lasa:
I have implemented Multiaddr in Elixir. 

- In Github: https://github.com/aratz-lasa/ex_multiaddr
- In Mix: https://hex.pm/packages/ex_multiaddr 

I do not know if I have to do something special or just say it here.
Does it need to be published on the official Multiformats repository (https://github.com/multiformats )? How does it work?


vmx:
Thanks!

PR #108, opened 2020-08-14
Add proxying protocols (SOCKS4/SOCKS4a, SOCKS5 and HTTP tunneling)

ntninja:
Needed for adding proxy support to py-ipfs-api-client.


ntninja:
@Stebalien: This one too, or no?

PR #109, opened 2020-08-24
Add TLS as per GH/multiformats/multicodec#145

ntninja:
This should be a no-brainer to merge as it was already merged in the MultiCodec repo and the two are supposed to be in sync with regards to MultiAddr formats.

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -19,11 +19,12 @@ code,	size,	name,	comment
 445,	296,	onion3,
 446,	V,	garlic64,
 447,	V,	garlic32,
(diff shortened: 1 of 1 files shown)


Stebalien (approved):
A bit late, but I see no real reason not to do this.

Concerns I've had before:

- Maybe we want it to take an argument?
- Are the semantics of HTTPs and /tls/http different?

In practice, I think the answer to both is "no". Unfortunately, authentication information will need to be inferred from the context given the... limitations of the multiaddr format.

PR #131, opened 2022-07-12
add webrtc

ckousik:
Add protocol code point for webrtc.
https://github.com/multiformats/go-multiaddr/pull/179
https://github.com/libp2p/specs/pull/412

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -33,5 +33,6 @@ code,	size,	name,	comment
 277,	0,	p2p-stardust,
 275,	0,	p2p-webrtc-star,
 276,	0,	p2p-webrtc-direct,
+280,	0,	webrtc, ICE-lite webrtc transport
 290,	0,	p2p-circuit,
 777,	V, memory, in memory transport for self-dialing and testing; arbitrary 


marten-seemann (approved):

protocols.csv
```diff
@@ -33,5 +33,6 @@ code,	size,	name,	comment
 277,	0,	p2p-stardust,
 275,	0,	p2p-webrtc-star,
 276,	0,	p2p-webrtc-direct,
+280,	0,	webrtc, ICE-lite webrtc transport
```
mxinden, line 36: If I understand correctly "ICE-lite" can only be used as a WebRTC server. In other words a browser can not do "ICE-lite", i.e. uses different UDP ports for each connection and does not know its port upfront and thus goes through the whole ICE process. Please correct me if I am wrong.

While I think the "ICE-lite webrtc transport" description is fine for now, I would hope for browsers to be able to re-use the same protocol component (`/webrtc`) in the browser-to-browser use-case, thus we might change **the description** later on.

mxinden (commented):
This looks good to me. Thanks @ckousik.

As far as I can tell `webrtc` and `certhash` is still missing in https://github.com/multiformats/multicodec/blob/master/table.csv. (Each multiaddr component is also a multicodec.) Would you mind filing a pull request there as well?

Once that is in, we can merge here.

mxinden:
//CC @melekes (no actions required)

protocols.csv
melekes: A lite implementation is only appropriate for devices that will *always* be connected to
the public Internet and have a public IP address at which it can receive packets from any
correspondent. ICE will not function when a lite implementation is placed behind a NAT
(RFC8445).

melekes (approved):

melekes:
opened https://github.com/multiformats/multicodec/pull/279

mxinden (approved):
> opened https://github.com/multiformats/multicodec/pull/279 and https://github.com/multiformats/multicodec/pull/280

With these merged, let's proceed here. Thanks @ckousik.

protocols.csv
mxinden: Just following up here, the initial hope, namely to reuse `/webrtc` for browser-to-browser (now called webrtc-w3c) is not possible. See https://github.com/libp2p/specs/pull/497 which uses `/webrtc-w3c` instead.

PR #134, opened 2022-08-21
Improve the regular expression

ben221199:
This pull request updates the regular expression of multiaddr.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ but only with multiaddr do we have a way of consistently addressing these networ
 
 ## Specification
 
-- Human-readable multiaddr: `(/<protoName string>/<value string>)+`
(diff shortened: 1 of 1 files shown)


marten-seemann:
I don't think this was intended as a regexp.

ben221199:
I don't know, but I do know that not every multiaddr segment holds a value, so the second part should be optional if you interpret it as RegEx.

marten-seemann:
The expression you wrote down is incorrect, if interpreted as a regexp. It doesn't even match the example given in the line below.

ben221199:
> The expression you wrote down is incorrect, if interpreted as a regexp. It doesn't even match the example given in the line below.

Okay, it is isn't fully RegEx, but because of the `()` as "group" and the `+` as "one or more", I thought the `?` as "optional" would make sense too.

PR #135, opened 2022-08-21
Add steps for encoding and decoding

ben221199:
This pull request adds steps to tell in a basic way how to encode and decode `multiaddr`s.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,11 +98,28 @@ Read more about it in [multiformats/unsigned-varint](https://github.com/multifor
 
 ### Encoding
 
-TODO: specify the encoding (byte-array to string) procedure
+1. Initiate an empty string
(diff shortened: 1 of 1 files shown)


mxinden (commented):
_I am not a maintainer of this repository. Thus feel free to ignore the below. I appreciate the help here and in the many other pull requests._

I am not sure how much value this pseudo code gives us. Would it not make more sense to point to one of the implementations, thus appointing that implementation as the reference implementation?

ben221199:
I understand your view about the pseudo-code. However, I think that it better to have some neutral pseudo-code instead of pointing developers coding in some language to an implementation in another language. I even think that the implementations in this GitHub organization shouldn't even be part of the specifications and just keep it only core level, but that is my point of view.

PR #137, opened 2022-08-21
Adding specification of more protocols

ben221199:
This pull request adds specifications of more protocols than only `dnsaddr`.

I think this pull request should be merged when I have added more, but now it is already possible to see the changes and talk about improvements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ and [protocols/](protocols/) for specifications of the currently supported proto
 
 TODO: most of these are way underspecified
 
-- /ip4, /ip6
+- [/ip4](protocols/IP4.md), /ip6
 - /ipcidr
 - /dns4, /dns6
 - [/dnsaddr](protocols/DNSADDR.md)
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ and [protocols/](protocols/) for specifications of the currently supported proto
 
 TODO: most of these are way underspecified
 
-- [/ip4](protocols/IP4.md), /ip6
+- [/ip4](protocols/IP4.md), [/ip6](protocols/IP6.md)
 - /ipcidr
 - /dns4, /dns6
 - [/dnsaddr](protocols/DNSADDR.md)
--- a/protocols/IP4.md
+++ b/protocols/IP4.md
@@ -1,0 +1,23 @@
+# `ip4`
+
+`ip4` is a protocol that defines which IPv4 address should be used.
+
+## Representation Format
+
+### Human-readable
+
+The human-readable format of the `ip4` protocol uses the quad-dotted format:
+
+	`127.0.9.1`
+
+TODO: Consider also supporting decimal format.
+	
+### Binary
+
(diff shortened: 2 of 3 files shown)


mxinden (commented):
Very much appreciated! Thanks for the help here!

To prevent any of the discussions on one protocol to block merging another, I would suggest to use separate pull requests for each. That is not a strong opinion though.

ben221199:
I think that adding the basis using this pull request is fine. We can add more information for a specific protocol using other pull requests.

@mxinden Also, can you tell me what the binary format of `dns`, `dns4`, `dns6` and `dnsaddr` is? Is it already defined? If not, I would suggest the FQDN format with a length-byte for every label.

Also, I have made some other pull requests on this and on other repositories. Hope someone can take a look at those too.

mxinden:
> @mxinden Also, can you tell me what the binary format of `dns`, `dns4`, `dns6` and `dnsaddr` is? Is it already defined? If not, I would suggest the FQDN format with a length-byte for every label.

Hope I understand your question correctly @ben221199. As far as I know it is just the utf-8 byte representation.

https://github.com/multiformats/rust-multiaddr/blob/911b18a51fafdee44a445dc491192f85430fbbbf/src/protocol.rs#L393-L416

ben221199:
```rs
w.write_all(encode::usize(bytes.len(), &mut encode::usize_buffer()))?;
```

This code above will prefix the data with an unsigned byte, yes.

Are the implementations in this organization authoritive at the moment, or is it still possible to alter the binary format at this time?

ben221199:
Okay, I think I added enough protocols for now. I think that, except for maybe some slight changes, this pull request can be merged.

mxinden:
> Are the implementations in this organization authoritive at the moment, or is it still possible to alter the binary format at this time?

They are used in many production systems, thus I would say yes.

protocols/DNS4.md
```diff
@@ -0,0 +1,21 @@
+# `dns4`
+
+`dns4` is a protocol that defines which domain name should be used, but only by resolving it to a IPv4 address.
```
mxinden, line 3:
```suggestion
`dns4` is a protocol that defines which domain name should be used, but only by resolving it to IPv4 addresses.
```

`dns4` can resolve to one or more ipv4 address.

protocols/P2P.md
```diff
@@ -0,0 +1,17 @@
+# `p2p`
+
+`p2p` is a protocol that defines which peer should be connected to using the PeerId containing its hashed public key.
+
+## Representation Format
+
+### Human-readable
+
+See the PeerId specification.
```
mxinden, line 9: How about linking to it here?

protocols/DNS4.md
ben221199: Good point. I will change that.

protocols/P2P.md
ben221199: That will be `https://docs.libp2p.io/concepts/peer-id/`, I guess.

ben221199:
Above suggestions are applied.

Issue #93, opened 2019-07-08
dnsaddr doesn't compose with p2p-circuit


Stebalien:
For example, I can't have:

```
/dnsaddr/foo.com/p2p/QmRelay/p2p-circuit/p2p/QmTarget`
```

This expects foo.com to resolve to `/something/something/p2p/QmTarget`.

Issue #87, opened 2019-02-05
Proposal: Add keyword arguments to protocols


alexander255:
# Problem description

The current MultiAddr spec does not have any good way for dealing with optional protocol parameters that have well defined defaults. Depending on the specific protocol in question different workarounds have been proposed with the predominant theme being *recursion*:

  * IPv6 link scopes: `/ip6/fe00::32/ip6zone/6/…`
  * TLS Server Name Identification: `/tls/sni/example.com/…`

This has the obvious problems that:

  * Each protocol must have a special parser then greedily swallow up all following components that it considers relevant
  * All possible of such “attribute protocol items” must be reserved to ensure that their names are no used as “regular” / “top-level” protocol names
     * As protocols evolve this may also cause nasty conflicts between newly defined attributes and existing protocol names.
     * There is no reason why attribute names may not be shared between different protocols but since they may never appear top-level protocol names they must still be treated as a separate class that shares a common namespace with the top-level class.
  * It is immediately obvious to a human reader which items are items of the previous top-level protocol and which constitute the start of a new encapsulation layer

There also do not seem any obvious advantages to this scheme that would somehow make the above problems reasonable trade-offs.

Another proposal suggested in some places (#63) was plain greediness: After a given protocol item shows up in the path, all further items are swallowed up and used as single “path parameter”:

  * HTTP: `/http/example.com/api/v1` (here `example.com` is the hostname and `/api/v1` the HTTP path)
  * WS and WSS: `/wss/example.com/api/v1/tls/ws`
  * Unix domain: `/unix/path/to/socket.sock/tls/ws`

While HTTP arguably is a terminator protocol (meaning that no other protocol may follow it anyways – this notion needs separate discussion!), Unix domain sockets and WebSockets definitely are not. Hence, it is unclear how a parser should figure out that `/tls` does not refer to a path component and this even is the case (the parser would have to proactively probe the file system for this, which is very much not in line with the vision of MultiAddr being a *common* description of paths to application endpoints; with WebSockets this is not even reliably possible to start with).

The example with WebSockets in particular demonstrates why this cannot work. A suggested alternative was to wrap the path parameter inside some kind of special set of delimiters (different kinds of braces were suggested):

  * `/wss/(/example.com/api/v1)/tls/ws`

While this works, it does not take into account the fact that there is nothing usually *required* about the given parameter: The hostname can usually be inferred from previous protocol levels (and left empty if unknown) and the path may always be empty.

[...]

Issue #67, opened 2018-06-12
/ws with paths not supported?


parkan:
It appears that specifying a path with websockets is not supported. This isn't explicit (the /protocols dir from the README is missing) but at all the examples I can find use a `/ip4/x.x.x.x/tcp/XXXX/ws` type format and the [test file for go-ipfs](https://github.com/multiformats/go-multiaddr/blob/92dbd0db4f2f05f40ce44df69ab89e3e2f3fda40/multiaddr_test.go#L20) specifies `/ws/foo` as an invalid test case.

e.g. the javascript library [ws](https://www.npmjs.com/package/ws) does allow you to specify a path:

```js
new WebSocket('ws://www.host.com/path')
```

while this isn't meaningful on the websockets side, it can be useful when e.g. using nginx to upgrade a connection to a particular path. Was this a conscious decision?

Stebalien:
This is probably due to the fact that multiaddrs suck at "path" arguments. They can sort of handle them as long as they appear at the end of the address unfortunately, we'd want to support `/ip4/x.x.x.x/tcp/XXXX/ws/a/b/c/ipfs/QmID`. Really, we'd need to write this as `/.../ws/[/a/b/c/d]/ipfs/QmID` (or some escaping mechanism). However, any time anyone brings up a proposal to deal with internal paths paths (e.g., multiaddrs as arguments to multiaddr protocols) the conversation starts at "the syntax is to complicated" and ends with "we don't really *need* them, do we?".

So, no, we don't currently support paths.

parkan:
got it, the current workaround is to use alternate ports with no path component -- works fine for us because we have complete firewall control, but that's not always the case

PR #140, opened 2022-09-08
Interpreting multiaddrs

MarcoPolo:
Adds a short section to the readme on how to interpret multiaddrs. I think this is the unifying principle of how we use multiaddrs in practice and could guide how we use them in the future. I don't think anything here is controversial and it fits with the "encapsulation" ideas presented originally. This is really just a clarification of what I think is the original intent.

This is a PR so that it's easier to discuss the changes line by line.

All but one of our constructions fit this principle. The uncommon `p2p-circuit` with a defined relay path does not fit this principle as the `p2p-circuit` component is currently defined, but I think we can make some small changes to make it fit. e.g. `/ip4/1.2.3.4/tcp/1234/p2p/qmR/p2p-circuit/ip4/4.3.2.1/tcp/54321/p2p/qmA` doesn't cleanly fit this model. We could add a new via protocol that takes in a single parameter of data, parses it as a multiaddr, and passes it to the underlying protocol as a `via` parameter. This would look like `/ip4/1.2.3.4/tcp/1234/p2p/qmR/p2p-circuit/via/bracketenc/[ip4/4.3.2.1/tcp/54321]/p2p/qmA`. What's nice is the could also describe (though we don't currently support) multiple hops in the circuit: `/ip4/1.2.3.4/tcp/1234/p2p/qmR/p2p-circuit/via/bracketenc/[ip4/4.3.2.1/tcp/54321/p2p/QmR2/p2p-circuit/via/bracketenc/[/ip4/2.2.2.2/tcp/1234]]/p2p/qmA`.

## Example 

As an example, here's how you would parse the string and interpret `/ip4/1.2.3.4/tcp/1234/p2p/qmR/p2p-circuit/via/bracketenc/[ip4/4.3.2.1/tcp/54321/p2p/QmR2/p2p-circuit/via/bracketenc/[/ip4/2.2.2.2/tcp/1234]]/p2p/qmA`. Remember that parsing the binary representation is easy since we either know the size of the data for a protocol (fixed size) or the data is length prefixed.

Parsing the string into a multiaddr, left to right:
1. Start with `/ip4/...` We see ip4 and call ip4's string to byte to consume the string up to `/tcp/...`.
2. We see `tcp` and call tcp's string to byte to consum up to `/p2p/...`.
3. continue to `via`, via doesn't take a value, so we continue to `bracketenc/...`
4. `bracketenc` parses the rest of the string until it finds a matching closing bracket `]`. This logic is defined by this encoding. The parsed string is kept as the value for `bracketenc`.
5. get to `p2p` and parse the following peer id.

Interpreting the multiaddr, right to left:
1. The `p2p` component takes its peer ID value and passes it to the next component as the parameter `peer=QmA`.
2. The `bracketenc` component takes its value and passes it to the next component as parameter `data=ip4/4.3.2.1/tcp/54321/p2p/QmR2/p2p-circuit/via/bracketenc/[/ip4/2.2.2.2/tcp/1234]`. It also forwards any other parameter it was given (i.e. `peer=QmA`).
3. The `via` component takes the data parameter and parses the multiaddr (similar to our parsing the string into a multiaddr example above). It passes this to the next component as parameter `via=<parsedMultiaddr>`. It also forwards any other parameter it was given (i.e. `peer=QmA`).
4. The `p2p-circuit` component now has the parameters `via=<parsedMultiaddr>` and `peer=QmA`. The `p2p-circuit` transport can now dial the relay `QmR` defined by the rest of the multiaddr (`/ip4/1.2.3.4/tcp/1234/p2p/qmR/`), and can tell that relay to connect to peer `QmA` _via_ the path `via=<parsedMultiaddr>`.
5. Done!

## Why now?

This came up as I was fixing WSS in go-libp2p and realized I needed to keep the hostname around for the SNI (e.g. I needed `/ip4/1.2.3.4/tls/sni/example.com/ws`).

We also need this to:

* Define a path to a websocket endpoint
* Passing SNI to a webtransport connection

## Related threads

https://github.com/multiformats/multiformats/pull/55. It's worth reviewing the `proposals/multipath.md` file first.

I think this is the minimal change that gets multiaddrs able to fulfill the goals of multipath. Key differences here are that we don't have a way to let the codec define the binary output directly (we always prefix it with a varint of byte length). We may want to change this and default it to varint prefix, but allow other ways to write and parse the binary. This would allow us to make the `p2p-circuit` fit the model in a backwards compatible way.

## Questions

- Is there anything wrong with this interpretation of multiaddrs?
- Is there something that multipath could provide that we don't get with this change?
- Can we keep this change as minimal as possible while still getting most of the benefits of multipath?

## Changes this would require

- This would involve deprecating the p2p-circuit + path multiaddr (which is relatively uncommon) in favor of the `p2p-circuit/via` or `p2p-circuit/route` option (as demonstrated above).
- This would involve (at least) changing how go-multiaddr parses string multiaddrs. It shouldn't be splitting on `/` instead it should parse the multiaddr from the left to the right. This should be backwards compatible, but new multiaddr constructions will not work on older clients. e.g. `/ip4/.../http/path/lenprefixencode/20_somepath/tosomething` wouldn't work on older clients, but this also wouldn't work because of the missing codec.

## What this enables/unlocks

If we agree on this we can unblock the following tricky issues:
This principle unblocks some previously tricky issues:
* https://github.com/multiformats/multiaddr/issues/67#issuecomment-396751334
* https://github.com/multiformats/multiaddr/issues/93
* https://github.com/multiformats/multiaddr/issues/87

We unlock the ability to handle parameters in a simple way. And we unlock the ability to support nested paths (via an encoding like: `/ip6/.../p2p/QmR1/p2p-circuit-path/lenprefixencode/33_/ip6/.../QmR2/p2p-circuit/p2p/QmA`)

---

Tagging folks I know have opinions here, forgive me if I miss you:
@Stebalien 
@marten-seemann 
@mxinden 
@elenaf9

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,44 @@ Multiaddr solves these problems by modelling network addresses as arbitrary enca
 
 Multiaddr was originally [thought up by @jbenet](https://github.com/jbenet/random-ideas/issues/11).
 
+## Interpreting multiaddrs
+
+Multiaddrs are parsed from left to right, but they should be interpreted right
+to left. Each component of a multiaddr wraps all the left components in its
+context. For example, the multiaddr `/dns4/example.com/tcp/1234/tls/ws/tls`
+(ignore the double encryption for now) is interpreted by taking the first `tls`
+component from the right and interpreting it as the libp2p security protocol to
+use for the connection, then passing the rest of the multiaddr to the websocket
+transport to create the websocket connection. The websocket transport sees
+`/dns4/example.com/tcp/1234/tls/ws/` and interprets the `tls` in this context to
+mean that this is going to be a secure websocket connection. The websocket
+transport also gets the host to dial along with the tcp port from the rest of
+the multiaddr.
+
+Components to the right can also provide parameters to components to the left,
+since they are in charge of the rest of the multiaddr's interpretation. For
+example, in `/ip4/1.2.3.4/tcp/1234/tls/p2p/QmFoo` the `p2p` component has the
+value of the peer id and it passes it to the next component, in this case the
+`tls` security protocol, as the expected peer id for this connection. Another
+example is `/ip4/.../p2p/QmR/p2p-circuit/p2p/QmA`, here `p2p/QmA` is passed to
+`p2p-circuit` and then the `p2p-circuit` component knows it needs to use the
+rest of the multiaddr as the information to connect to the relay node.
+
+This enables nesting and arbitrary parameters. A component can parse
+arbitrary data with some encoding and pass it as a parameter to the next
+component of the multiaddr. For example, we could reference a specific HTTP path
+by composing `path` and `urlencode` components along with an `http` component.
+This would look like
+`/dns4/example.com/http/path/percentencode/somepath%2ftosomething`. The
+`percentencode` parses the data and passes it as a parameter to `path`, which
+passes it as a named parameter (`path=somepath/tosomething`). A user may not
+like percentencode for their use case and may prefer to use `lenprefixencode` to
+have the multiaddr instead look like
+`/dns4/example.com/http/path/lenprefixencode/20_somepath/tosomething`. This
+would work the same and require no changes to the `path` or `http` component.
+It's important to note that the binary representation of the data in
+`percentencode` and `lenprefixencode` would be the same. The only difference is
+how it appears in the human-readable representation.
 
 ## Use cases
 


mxinden (commented):
From a high level, this looks good to me. rust-libp2p follows the convention of interpreting a multiaddress from right to left.

I don't think I am deep enough in the matter to have an opinion. I will let others comment first. In case others feel the same or don't have the time to comment, I will give this more thought. If so, please ping me @MarcoPolo.

MarcoPolo:
Had a sync meeting with @Stebalien and here's where we ended up:

* Steven is in general agreement with this interpretation strategy.
* Would suggest using bracket encoding with a percent escape mechanism. There is some set of allowed characters, and everything else would be percent encoded escaped. e.g. a path would look like: `/dns4/example.com/http/path/bracketenc/[foo/bar/baz]` or, if the path had a bracket in it, ` /dns4/example.com/http/path/bracketenc/[foo/bar%5D/baz]` (%5D is the percent encoding of ]). This has the benefit of looking nice for many usecases. Ultimately the binary representation is the same, so it's not worth bikeshedding too much here.
* Suggest starting with a single encoding strategy to minimize the amount of changes required by implementations.

MarcoPolo:
Moving this out of draft so we can further discuss before potentially merging.

Keep in mind this change is not meant to define what encoding we should choose, only that we should interpret multiaddrs right to left (and parse them left to right).

mxinden (approved):
Having given this more thought, this looks good to me. Unless someone objects, I am in favor of merging.

PR #142, opened 2022-09-28
Check if multicodec table is in sync with the multiaddr table

MarcoPolo:
## Summary

Adds a simple check to make sure codecs defined in the multiaddr table are in sync with the multicodec table. See the github workflow run for this PR as an example.

## Before Merge

- [x] Allow at least 24 hours for community input

### DIFF
--- a/.github/workflows/check-multicodec-in-sync.yml
+++ b/.github/workflows/check-multicodec-in-sync.yml
@@ -1,0 +1,22 @@
+name: Check that we are in sync with the Multicodec repo
+on: [pull_request]
+jobs:
+  check-multicodec-in-sync:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - name: Checkout tools repo
+        uses: actions/checkout@v3
+        with:
+          repository: multiformats/multicodec
+          path: multicodec
+      - name: List files in the repository
+        run: |
+          ls ${{ github.workspace }}
+      - name: Install babashka
+        uses: just-sultanov/setup-babashka@v2
+        with:
+          version: '0.4.6'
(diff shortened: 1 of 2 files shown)


MarcoPolo:
https://github.com/multiformats/multicodec/pull/294 should fix the failing workflow.

MarcoPolo:
> This test seems useful. I’m a bit hesitant since I’m not familiar with Clojure, and this code could become hard to maintain.

Would you prefer bash instead ;) ?

Joking aside, I considered doing this in many other languages, but found the Clojure version more concise and simpler (and more fun!). The code is 60 lines, and there's no esoteric features here. However, if you want to write the Go version, feel free!

marten-seemann:
> Would you prefer bash instead ;) ?

This could be a sharness test! 🤪

marten-seemann (approved):

MarcoPolo:
CI now passes after https://github.com/multiformats/multicodec/pull/294 was merged

BigLep:
> This test seems useful. I’m a bit hesitant since I’m not familiar with Clojure, and this code could become hard to maintain.

I think this is an important point.  I'm worried about us introducing new languages without it being deliberate.  JS or Python seem more likely for new folks to understand than Clojure.

BigLep:
Thanks for sharing @MarcoPolo - the reasoning makes sense. 

Completely agreed about not using Bash.

I'm less-worried about this specific instance given this is not an area where we're we expect to be adding a lot more scripting.  

My biggest worry is about the precedent as I want to make sure that introducing new languages and technologies is very deliberate and thought through from a long-term view.  

I don't think we need to discuss more now, as long as we have the takeaway for an upfront conversation for the next time, especially if it is in an area we expect to build on or expand.

Thanks.

PR #144, opened 2022-10-13
Add host-header to protocol.csv

thibmeu:
Host header is required for HTTP based protocol (HTTP, HTTPS, WS, WSS) when DNS(, 4, 6, addr) has been resolved to an IP address and TCP port.

This is required to preserve the host header while resolving a DNS address. The tuple ip:port alone is not sufficient to contact a host via HTTP.

For instance, this is going to be used to address libp2p/go-libp2p#1829

cc @MarcoPolo

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -26,6 +26,7 @@ code,	size,	name,	comment
 460,	0,	quic,
 465,	0,	webtransport,
 466,    V,	certhash,
+481,  V,  host-header, HTTP host header
 480,	0,	http,	HyperText Transfer Protocol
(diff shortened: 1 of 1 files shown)


MarcoPolo:
Some discussion around the number and the name here: https://github.com/multiformats/go-multiaddr/pull/185#discussion_r994610001

thibmeu:
changed code from `476` to `481`

MarcoPolo:
Closing this for now, since we ended up just using SNI for this as well. This may come up again in the future though.

PR #147, opened 2022-10-28
add a new code point for QUIC v1

marten-seemann:
To be merged after https://github.com/multiformats/multicodec/pull/298.

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -23,10 +23,10 @@ code,	size,	name,	comment
 448,	0,	tls,	Transport Layer Security
 449,	V,	sni,	Server Name Indication RFC 6066 § 3
 454,	0,	noise,
-460,  0,	quic,
-461,  0,	quic-v1,
+460,	0,	quic,
+461,	0,	quic-v1,
 465,	0,	webtransport,
-466,  V,	certhash,
+466,	V,	certhash,
 480,	0,	http,	HyperText Transfer Protocol
 443,	0,	https,	Deprecated alias for /tls/http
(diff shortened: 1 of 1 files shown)


protocols.csv
```diff
@@ -23,9 +23,10 @@ code,	size,	name,	comment
 448,	0,	tls,	Transport Layer Security
 449,	V,	sni,	Server Name Indication RFC 6066 § 3
 454,	0,	noise,
-460,	0,	quic,
+460,  0,	quic-draft29,
```
MarcoPolo, line 26: Let's not make this rename change yet, because I think it may break parsing existing strings.

MarcoPolo (changes requested):

protocols.csv
marten-seemann: Reverted this change.

protocols.csv
```diff
@@ -23,9 +23,10 @@ code,	size,	name,	comment
 448,	0,	tls,	Transport Layer Security
 449,	V,	sni,	Server Name Indication RFC 6066 § 3
 454,	0,	noise,
-460,	0,	quic,
+460,  0,	quic,
```
MarcoPolo, line 26: Did the formatting break?

MarcoPolo (approved):

protocols.csv
marten-seemann: Yes. Editors break it all the time, since we're using tabs, and many editors convert those to spaces (including the built-in GitHub editor).

PR #151, opened 2023-03-09
fix: Rename webrtc-w3c to webrtc-private-to-private

mxinden:
## Summary

We have decided to rename `webrtc-w3c` to `webrtc-private-to-private`. See rational in https://github.com/multiformats/multiaddr/pull/150/#issuecomment-1460912728.

No software has been released using `webrtc-w3c`, thus renaming is not a breaking change.

## Before Merge

  - See https://github.com/multiformats/multicodec/pull/318

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -36,6 +36,6 @@ code,	size,	name,	comment
 275,	0,	p2p-webrtc-star,
 276,	0,	p2p-webrtc-direct,
 280,	0,	webrtc, ICE-lite webrtc transport
-281, 	0,	webrtc-w3c, webrtc transport where connection establishment is according to w3c spec
+281, 	0,	webrtc-private-to-private, WebRTC transport establishing connection between two private nodes
 290,	0,	p2p-circuit,
 777,	V, memory, in memory transport for self-dialing and testing; arbitrary 


mxinden:
Friendly ping @achingbrain, @MarcoPolo and @marten-seemann.

MarcoPolo (approved):

achingbrain:
Cross posting my comment from [here](https://github.com/multiformats/multiaddr/pull/150#issuecomment-1467445889) as it's a closed issue:

> Would it be fair to say that what we're calling `/webrtc-private-to-private`, if we can't call it `/webrtc-w3c` (which, on balance, is not a completely terrible name since it uses the w3c flow and the spec mentions the w3c APIs specifically) is really something like `/webrtc+w3c` or `/webrtc+w3c-flow` or something, using a `+` to differentiate between the transport and our connection protocol we place on top of it, and what we're calling `/webrtc` is really `/webrtc+sdp-munging`?

I realise it's the 11th hour, but it seems that `/webrtc-private-to-private` is not a transport in the same way that `/tcp` or `/ws` is a transport - the transport is webrtc and `private-to-private` is a way of using that transport.  (an aside: I'm not sure what `private` means in this context?  Vs `public`?  Undiallable?  Webrtc makes browser nodes diallable. I am at a loss.)

The browser-to-server `/webrtc` multiaddr segment name seems to imply there's nothing special or magic about it, so a fresh pair of eyes may assume you can implement it with any spec-compliant webrtc implementation, but this is not the case as it uses sdp munging which is our own spec [notes is disallowed by the webrtc spec](https://github.com/libp2p/specs/tree/master/webrtc) and as such is unimplementable in any spec-compliant environment.

That's fine and all, but the unadorned `/webrtc` seems a bit misleading.

protocols.csv
```diff
@@ -36,6 +36,6 @@ code,	size,	name,	comment
 275,	0,	p2p-webrtc-star,
 276,	0,	p2p-webrtc-direct,
 280,	0,	webrtc, ICE-lite webrtc transport
```
achingbrain, line 38:
```suggestion
280,	0,	webrtc+sdp-munging, ICE-lite WebRTC transport with SDP munging
```

protocols.csv
```diff
@@ -36,6 +36,6 @@ code,	size,	name,	comment
...
-281, 	0,	webrtc-w3c, webrtc transport where connection establishment is according to w3c spec
+281, 	0,	webrtc-private-to-private, WebRTC transport establishing connection between two private nodes
```
achingbrain, line 39:
```suggestion
281, 	0,	webrtc, WebRTC transport establishing connection between two WebRTC spec compliant nodes
```

achingbrain (changes requested):
Not that I want to put words in @marten-seemann's mouth, but from a different channel:

> *achingbrain* So you'd be happy calling browser to browser /webrtc and browser to server /webrtc+sdp-munging?
> *Marten Seemann* Yes. I actually like calling out that we’re doing the munging
> Because that’s the crucial difference between the two

This makes it clear that a spec compliant WebRTC implementation can use this address to communicate, and ones that are more flexible can let peers know that they can use SDP munging via a different address.

achingbrain:
Closing in favour of https://github.com/multiformats/multiaddr/pull/152

PR #164, opened 2023-08-24
Define `/httppath`

MarcoPolo:
## Summary

Adds a new multiaddr component for `/httppath`, and define it.

I think this follows RFC 3986's notion of "Path" (section [3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3)). Please correct me if I'm wrong.

I'm using `httppath` instead of `httpath` (used by IPNI) previously for two reasons:

1. I prefer joining two full words.
2. In case there's any conflicting interpretation, this will be unambiguous as it has a different name. 

## Before Merge

### DIFF
--- a/protocols/httppath.md
+++ b/protocols/httppath.md
@@ -1,0 +1,40 @@
+# `httppath`
+
+This protocol encodes an HTTP Path to a resource. In the string representation,
+characters in the reserved set are percent-encoded ( "/" becomes "%2F").
+Percent-encoding itself is defined by [RFC 3986 Section
+2.1](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1). In the binary representation, no escaping is needed as the value is length prefixed.
+
+To ease implementation and benefit from reusing existing percent-encoding logic
+present in many environments (Go's
+[url.PathEscape](https://pkg.go.dev/net/url@go1.21.0#PathEscape); JS's
+[encodeURIComponent](http://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent);
+and Rust's [many
+crates](https://crates.io/search?q=percent%20encode&sort=downloads)), it is
+acceptable to encode more characters than the reserved set. While it's not
+necessary to encode a space " " as %20, it's is acceptable to do so.
+
+When comparing multiaddrs, implementations should compare their binary
+representation to avoid ambiguities over which characters were escaped.
+
+## Reserved Characters
+
+| Character | Reason                        |
+| --------- | ----------------------------- |
+| /         | Multiaddr component separator |
+| %         | Percent encoding indicator    |
+| ?         | Marks the end of an HTTP path |
+
+## Usage
+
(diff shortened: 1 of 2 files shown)


Winterhuman:
Out of curiosity, are the paths strictly ASCII characters (with punycode needed for UTF-8 conversion), or is encoding the paths as UTF-8 allowed?

MarcoPolo:
Good question. [Section 2 of RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2) covers this. Basically it depends on the context. If the context/protocol is UTF-8, then this is UTF-8. In practice I think the string encoding of this will be UTF-8.

MarcoPolo:
I think as is, my document is ambiguous. Only the following characters are not percent-encoded: `unreserved / pct-encoded / sub-delims / ":" / "@"`. 

I was trying to avoid the ABNF notation from the RFC, but I might have to include it.

protocols/httppath.md
```diff
@@ -0,0 +1,43 @@
...
+`/httppath` should be appended to the end of an existing multiaddr, including after the peer id component (p2p). As an example, here's a multiaddr referencing the `.well-known/libp2p` HTTP resource along with a way to reach that peer:
+
+```
+/ip4/1.2.3.4/tcp/443/tls/http/p2p/12D.../httppath/.well-known%2Flibp2p
+```
+
+The `/httppath` component can also be appended to just the `/p2p/...` component, and rely on a separate peer discovery mechanism to actually identify the peer's address:
+
+```
+/ip4/1.2.3.4/tcp/443/tls/http/p2p/12D.../httppath/.well-known%2Flibp2p
```
thomaseizinger, line 42:
```suggestion
/p2p/12D.../httppath/.well-known%2Flibp2p
```

Did you mean to not include the address here?

achingbrain:
It's an aesthetic point but `/httppath` is quite hard to read.

I wonder if it could be `/http-path` (like `/p2p-circuit` or `/webrtc-direct`) or even `/http+path`, since it doesn't really work without `/http` and makes that section of the multiaddr more specific?

It also opens the door for other modifiers that might be required to create a valid request, such as `/http+host/example.com`, `/http+header/foo=bar`, `/http+auth/user:password`, `/http+cookie/foo=bar` etc.

MarcoPolo:
`/http-path` works for me. 👍

achingbrain:
Is a multiaddr with `/http-path` valid without `/http` also being present?

If not the spec should say so.  If so, it should say how to infer a `/http` tuple when it's not present.

MarcoPolo:
> Is a multiaddr with /http-path valid without /http also being present?

Yes. There's an example in the spec with just a `/p2p` component.

`/http` refers to the other end supporting an HTTP transport. The `/http` component is defined at https://github.com/libp2p/specs/blob/master/http/transport-component.md.

achingbrain (approved):
LGTM

MarcoPolo:
I don't expect this component to be present in multiaddrs that a peer shares for connectivity as in Peer Routing Records. Only in cases where you want to reference a specific HTTP resource. 

I think DNS records are a good analogy. For example, An `A` query gives you only ipv4 address to a domain name. It just gives you the thing you need to find the server but doesn't give you anything else. Our routing records should be similar. Just the information to reach a server and its peer id.

PR #172, opened 2024-06-05
chore: Fill in todos

MarcoPolo:
## Summary
Unglamarous. Just filling in the missing details of encoding/decoding

## Before Merge

- [x] If this is a new protocol, has the change been applied to https://github.com/multiformats/multicodec as well? If so, please link the multicodec PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,16 +7,18 @@
 
 > Composable and future-proof network addresses
 
-- [Introduction](#introduction)
-- [Use cases](#use-cases)
-  - [Encapsulation based on context](#encapsulation-based-on-context)
-- [Specification](#specification)
-  - [Encoding](#encoding)
-  - [Decoding](#decoding)
-- [Protocols](#protocols)
-- [Implementations](#implementations)
-- [Contribute](#contribute)
-- [License](#license)
+- [multiaddr](#multiaddr)
+  - [Introduction](#introduction)
+  - [Interpreting multiaddrs](#interpreting-multiaddrs)
+  - [Use cases](#use-cases)
+    - [Encapsulation based on context](#encapsulation-based-on-context)
+  - [Specification](#specification)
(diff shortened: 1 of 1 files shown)


Dustin4444 (approved):

README.md
```diff
@@ -132,13 +134,33 @@ Multiaddr and all other multiformats use unsigned varints (uvarint).
 Read more about it in [multiformats/unsigned-varint](https://github.com/multiformats/unsigned-varint).
 
 
-### Encoding
+### Encoding (Bytes -> String)
 
-TODO: specify the encoding (byte-array to string) procedure
+1. Read the first byte and find the corresponding protocol.
```
achingbrain: Since varint-encoding is mentioned in the decoding section, it's probably worth mentioning it here in case the reader thinks all you need to do is read the first byte and you have a protocol code.

```suggestion
1. Varint-decode the first byte(s) and find the corresponding protocol.
```

README.md
```diff
@@ -132,13 +134,33 @@ Multiaddr and all other multiformats use unsigned varints (uvarint).
 Read more about it in [multiformats/unsigned-varint](https://github.com/multiformats/unsigned-varint).
 
 
-### Encoding
+### Encoding (Bytes -> String)
```
achingbrain, line 137: I know this has been in the spec for like, ever, but is this the right way round?  IPLD generally refers to "encoding" as turning an object (string, here) into bytes and "decoding" as getting an object from bytes.

MarcoPolo: wdyt about `String Encoding` instead?

achingbrain: I think it has the same problem since "encoding" is still in the title.

This README was written a long time ago and the ecosystem has coalesced around certain language and conventions since then, I think we should bring this in line with everything else.

Issue #139, opened 2022-09-08
UNIX addresses can be ambiguous.


saul-avikar:
Unix domain socket addresses become ambiguous when they encapsulate other addresses. This is obviously because unix addresses are not a fixed length and can contain forward slashes. This becomes an issue when you need to when you need to encapsulate other protocols like `/p2p` for example:
```
/unix//tmp/node.socket/p2p/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ
```
This type of address would be needed if you have multiple nodes on the same machine that need to connect to each other. (VMs/Docker comes to mind.)

This also relates to issue: #55

littlebenlittle:
This is also useful for clients that can connect to both local and remote services. For example, it would not be unreasonable to run the IPFS API over a unix socket. However you may also want to use the IPFS client to manage a remote API. Right now these cases have to be handled separately.

As was [pointed out](https://github.com/multiformats/multiaddr/issues/63#issuecomment-546289697) in #63 by infinity0, unix domain sockets can be either [stream or datagram](https://man7.org/linux/man-pages/man7/unix.7.html). Therefore I recommend a multiaddr format such as

```
/unix/stream/(/path/to/sock)/some-protocol/0.1.0
or
/unix/datagram/(/path/to/sock)/some-protocol/0.1.0
```

It's not pretty, but it's a path forward for a problem that's been around for several years.

diegomrsantos:
Another user case is for Tor transport. We can start a peer with`/unix/xxx/onion3/xxx`. The unix socket is the address configured as a onion service in Tor and the onion address is the one used to dial the peer.

@littlebenlittle we could have a shorter version `/unix/s/xxx` and  `/unix/d/xxx`. One inconvenience is that `\path\to\sock` is a valid path name on unix systems.

Issue #55, opened 2017-11-06
Protocol specifiers are ambiguous?


icefoxen:
If you use Unix file paths and delimiters, it seems like it would be really easy to make things ambiguous.

First off, I don't see a "file" protocol type, is this the the `unix` protocol type listed in protocols.csv?

If so, what happens if I make a file path `/home/icefox/tcp/127.0.0.1/foo` and try to represent it as a multiaddr path?

```
/unix/home/icefox/tcp/127.0.0.1/foo`
```

Okay, that's a little artificial, and you could specify that you can disambiguate from context; a `unix` protocol address cannot contain a `tcp` protocol subaddress, though that seems to defeat the purpose; things stop being self-describing and become context-dependent.  (And on Plan9 you certainly can do TCP via the filesystem.)

Okay, what about a more concrete example using already defined things:

```
/wss/example.com/http/127.0.0.1/
```

Is this accessing the websockets endpoint `wss://example.com/http/127.0.0.1`, or is it trying to nest the HTTP protocol inside the websockets endpoint `wss://example.com/` ?

alexander255:
(Yes, I know this has been open for quite some time, but since nobody responded…)

I think you were being misled here:
MultiAddr is not about describing paths to *data*, but rather about describing paths to *endpoints* (applications on hosts). Your example of trying to describe a file path (/home/icefox/tcp/127.0.0.1/foo) therefor makes no sense in the context MuiltAddr, since it simply cannot describe these kinds of things. There is also no application you need to *talk to* to get access to that file hence there is no need for MultAddr at all.
(You could argue that you are still talking to the kernel hence on could argue that `/kernel` or `/kernel-vfs` or maybe also just `/file` may make sense, but that's not really what you meant to say.)  The `/unix` protocol refers to [Unix domain sockets](https://en.wikipedia.org/wiki/Unix_domain_socket), not UnixFS-paths by the way.

[...]

PR #174, opened 2024-10-28
Define `/unix`

achingbrain:
## Summary

Adds a protocol note for how to encode paths to Unix domain sockets as strings, that may include the delimiting character of `/`.

This allows us to append other tuples to the multiaddr while also ensuring we can round-trip the address to a string and back.

This doesn't affect the binary representation of the multiaddr since everything is length-delimited.

Takes inspiration from #164 and proposes using URI encoding for the segment, the same as the `/http-path` component.

One difference is if the path is to represent the filesystem root, it must be included in the value portion of the tuple, otherwise it can be omitted.

## Before Merge

- [x] Allow at least 24 hours for community input

### DIFF
--- a/protocols/unix.md
+++ b/protocols/unix.md
@@ -1,0 +1,36 @@
+# `unix`
+
+This protocol encodes a Unix domain socket path to a resource. In the string
+representation, the path is encoded in a way consistent with a single URI Path
+segment per [RFC 3986 Section 3.3](https://datatracker.ietf.org/doc/html/rfc3986#autoid-23).
+
+Specifically following the grammar of a single `segment-nz`. In the binary
+representation, no encoding is needed as the value is length prefixed.
+
+When comparing multiaddrs, implementations should compare their binary
+representation to avoid ambiguities over which characters were escaped.
+
+## Examples
+
+The following is a table of examples converting some common Unix paths to their
+Multiaddr string form.
+
+| Unix Path                   | Multiaddr string form                   |
+| --------------------------- | --------------------------------------- |
+| /                           | `/unix/%2F`                             |
+| /file.socket                | `/unix/file.socket`                     |
+| /dir/file.socket            | `/unix/dir%2Ffile.socket`               |
+| /dir/file.socket/p2p/12D... | `/unix/dir%2Ffile.socket/p2p/12D...`    |
+| /tmp/foo/../bar             | `/unix/tmp%2Ffoo%2F..%2Fbar`            |
+| /%2F                        | `/unix/%252F`                           |
+| a%20space                   | `/unix/a%2520space`                     |
+| a%2Fslash                   | `/unix/a%252Fslash`                     |
+
+## Usage
+
+`/unix` would typically be found at the start of a multiaddr, however it may
+appear anywhere, for example in the case where we route through some sort of
+proxy server or SSH tunnel.
+
+The leading `/` character of the path can be omitted, unless it is the only
+character in the path, in which case it must be escaped as normal.
--- a/protocols/unix.md
+++ b/protocols/unix.md
@@ -23,8 +23,8 @@ Multiaddr string form.
 | /dir/file.socket/p2p/12D... | `/unix/dir%2Ffile.socket/p2p/12D...`    |
 | /tmp/foo/../bar             | `/unix/tmp%2Ffoo%2F..%2Fbar`            |
 | /%2F                        | `/unix/%252F`                           |
-| a%20space                   | `/unix/a%2520space`                     |
-| a%2Fslash                   | `/unix/a%252Fslash`                     |
+| /a%20space                  | `/unix/a%2520space`                     |
+| /a%2Fslash                  | `/unix/a%252Fslash`                     |
 
 ## Usage
 
--- a/protocols/unix.md
+++ b/protocols/unix.md
@@ -18,13 +18,14 @@ Multiaddr string form.
 | Unix Path                   | Multiaddr string form                   |
 | --------------------------- | --------------------------------------- |
 | /                           | `/unix/%2F`                             |
-| /file.socket                | `/unix/file.socket`                     |
-| /dir/file.socket            | `/unix/dir%2Ffile.socket`               |
-| /dir/file.socket/p2p/12D... | `/unix/dir%2Ffile.socket/p2p/12D...`    |
-| /tmp/foo/../bar             | `/unix/tmp%2Ffoo%2F..%2Fbar`            |
+| /file.socket                | `/unix/%2Ffile.socket`                  |
+| /dir/file.socket            | `/unix/%2Fdir%2Ffile.socket`            |
+| /dir/file.socket/p2p/12D... | `/unix/%2Fdir%2Ffile.socket/p2p/12D...` |
+| /tmp/foo/../bar             | `/unix/%2Ftmp%2Ffoo%2F..%2Fbar`         |
 | /%2F                        | `/unix/%252F`                           |
-| /a%20space                  | `/unix/a%2520space`                     |
-| /a%2Fslash                  | `/unix/a%252Fslash`                     |
+| /a%20space                  | `/unix/%2Fa%2520space`                  |
+| /a%2Fslash                  | `/unix/%2Fa%252Fslash`                  |
+| socket                      | `/unix/socket`                          |
 
 ## Usage
 
@@ -32,5 +33,5 @@ Multiaddr string form.
 appear anywhere, for example in the case where we route through some sort of
 proxy server or SSH tunnel.
 
-The leading `/` character of the path can be omitted, unless it is the only
-character in the path, in which case it must be escaped as normal.
+The absence of a `/` character at the start of the decoded address indicates a
+relative path, otherwise the path is absolute.
--- a/protocols.csv
+++ b/protocols.csv
@@ -13,7 +13,7 @@ code,	size,	name,	comment
 132,	16,	sctp,
 301,	0,	udt,
 302,	0,	utp,
-400,	V,	unix,
+400,	V,	unix,	Percent-encoded path to a Unix domain socket
 421,	V,	p2p,	preferred over /ipfs
 421,	V,	ipfs,	backwards compatibility; equivalent to /p2p
 444,	96,	onion,
@@ -39,4 +39,4 @@ code,	size,	name,	comment
 280,	0,	webrtc-direct, ICE-lite webrtc transport with SDP munging during connection establishment and without use of a STUN server
 281, 	0,	webrtc, webrtc transport where connection establishment is according to w3c spec
 290,	0,	p2p-circuit,
-777,	V, memory, in memory transport for self-dialing and testing; arbitrary 
+777,	V, memory, in memory transport for self-dialing and testing; arbitrary


protocols/unix.md
```diff
@@ -0,0 +1,36 @@
...
+| Unix Path                   | Multiaddr string form                   |
+| --------------------------- | --------------------------------------- |
+| /                           | `/unix/%2F`                             |
+| /file.socket                | `/unix/file.socket`                     |
+| /dir/file.socket            | `/unix/dir%2Ffile.socket`               |
+| /dir/file.socket/p2p/12D... | `/unix/dir%2Ffile.socket/p2p/12D...`    |
+| /tmp/foo/../bar             | `/unix/tmp%2Ffoo%2F..%2Fbar`            |
+| /%2F                        | `/unix/%252F`                           |
+| a%20space                   | `/unix/a%2520space`                     |
+| a%2Fslash                   | `/unix/a%252Fslash`                     |
```
achingbrain, line 27:
```suggestion
| /a%20space                  | `/unix/a%2520space`                     |
| /a%2Fslash                  | `/unix/a%252Fslash`                     |
```

MarcoPolo:
Are there any backwards compatibility issues? Is anyone using this multiaddr currently?

aschmahmann:
> Is anyone using this multiaddr currently

I'm not sure how heavily it's utilized, but kubo can use those addresses. e.g. https://github.com/ipfs/kubo/blob/4009ad3e5a502518ddc7d48a888707f812ddc629/docs/config.md?plain=1#L219

aschmahmann:
> Are there any backwards compatibility issues?

I haven't looked too deeply at usage, but I'd suspect so. The question is what to do about it. There are a bunch of issues that are basically about the failures related to protocols (like http or unix) in taking unescaped paths like  #139  #55. They were also known from the earliest days https://github.com/multiformats/go-multiaddr/pull/31.

Maybe the answer is to just break it, or to say that implementations MAY/SHOULD/MUST preferentially try looking for the entire path as a socket and if that fails will try with just the first path segment as a socket.

----

Not specific to this PR, but between http-path and this perhaps worth biting the bullet on specifying how "path" types should be done in general. Maybe the answer is just to require them all to be escaped, maybe it's something else (e.g. closer to @Stebalien's proposal in https://github.com/multiformats/multiformats/pull/55).

achingbrain:
There will be backwards compatibility issues with the string representation of the path due to the escaping, though not with the binary representation.

I think this is used but not very commonly.  The js stack can also use unix addresses but currently only as a terminal element, that is no tuples can follow it.

Arguably the lack of escaping has harmed it's use - I know I've tried to use it a few times over the years but always come up hard against the inability to append anything else to the address and the various issues asking for clarification etc speak to a need for this.

> or to say that implementations MAY/SHOULD/MUST preferentially try looking for the entire path as a socket and if that fails will try with just the first path segment as a socket

I wonder if this could get exploited by making longer paths with segments that mirror tuples after the unix section?  Possibly with `..` in them that go somewhere bad?

> Maybe the answer is to just break it

I kind of think so.  The [original PR](https://github.com/multiformats/go-multiaddr/pull/31) was merged in haste, the question about [if we need to append things to it](https://github.com/multiformats/go-multiaddr/pull/31#issuecomment-246333387) was [unresolved](https://github.com/multiformats/go-multiaddr/pull/31#issuecomment-246513565).

It was flagged as experimental and subject to change so..

> proposal in https://github.com/multiformats/multiformats/pull/55

It's interesting that - it seems to mandate parsing paths left to right, we recently switched multiaddr-to-uri to parsing [right to left](https://github.com/multiformats/js-multiaddr-to-uri/pull/120) to support some forms of multiaddr.

It also doesn't seem to say much about escaping so we'd probably still need to solve the same problem there.

achingbrain:
Are there any further thoughts here?

ntninja:
@achingbrain: Since I just noticed it wasn’t mentioned I'd just like to mention my old MultiAddr proposal from 2019 again: #87 

It’s mainly about argument to individual MultiAddr path segments, but it also proposes syntax for encoding paths in MultiAddr segments:
 * `/unix/(/dir/file.socket)/http`, rather than `/unix/dir%2Ffile.socket/http`
 * `/unix/(/dir\\weird/\(file\).sock)/ws`, rather than `/unix/dir\weird%2F(file).sock/ws`
 * `/unix/dir/file.socket`, rather than `/unix/dir%2Ffile.socket` for backward-compatibility

This is IMHO significantly easier to read than using percent-encoded local file system paths.

achingbrain:
Thanks, I didn't see that issue.  I like your proposal for using `()` for tuple values and the flexibility it brings but I think it's more in the realm of a Multiaddr v2 as you point out.  I would certainly support that style of stringification if it those sort of discussions started happening.

To solve this particular issue though, I would prefer to be consistent with other multiaddr tuple types.  HTTP has already solved this problem by percent escaping the characters so I think we should use that here too.

ntninja:
@achingbrain: I can see where you’re coming from, but I’d just like the point out that the major benefit, besides nicer style, is the fact that it’s backwards-compatible! (We *are* talking about MultiAddr v1 here after all. :slightly_smiling_face:)

I know most of the MultiFormat people tend to lean on the string-representation => unimportant side, but at least `ipfs-cluster` and `py-ipfs-http-client` are using the `/unix/dir/file.socket` *string-representation* as part of their (potentially) persistent configuration right now. So changing it to `/unix/dir%2Ffile.socket` would actually be a backward-incompatible/major-release change for them.

By contrast, `/unix/(/dir/file.socket)/…` would remain compatible except in the super-obscure edge-case of a socket file path actually starting with `/(/`…

Not saying you shouldn’t go with your proposal, but that *is* something to thoroughly consider I think!

dhuseby:
Are there any limitations on the character set that can be used in a path? I'm assuming that the characters are all UTF-8 encoded so unicode characters like ↑ (e.g. up arrow, UTF-8: `0xE2 0x86 0x91`) are encoded as `%25E28691`. Is that correct? Or would it be encoded as `%E28691`?

achingbrain:
> Are there any limitations on the character set that can be used in a path?

Not really, it's kind of the wild west since most "unix" implementations can trace their lineage prior to UTF-8 and other modern conveniences.

POSIX has a definition of [portable file name characters](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282) but realistically it could be UTF-8, ASCII or char[].  I think just not NUL `\0`

As per this proposal `↑` would be encoded as `%E2%86%91`.

Full details are in [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), linked to from the proposed changes in the PR.

achingbrain:
Refs https://github.com/ipshipyard/roadmaps/issues/16

MarcoPolo (approved):
Commenting on the spec itself, it seems reasonable. I won't comment on whether this is safe to deploy or if it will break existing users of the unix multiaddr component.

achingbrain:
> Allow at least 24 hours for community input

Given this has been open for over six months I think this has been satisfied.

I will merge this next week unless there are strenuous objections.

nabijaczleweli:
*If* this only governs the serialised form: Please do *not* merge this while either (a) this implies that an initial / is inserted if a path doesn't have one or (b) this implies that a relative path somehow magically becomes an absolute one. These are both severe footguns, and in the (a) case it also *excludes* serialising abstract socketaddrs.

*If* this governs both the serialised form *and* opines on ORM/binary forms: Please do *not* merge this. The spec is *not* reasonable because it (a) doesn't include abstract socketaddrs, (b) allows dropping the initial / which is nonsensical since that, instead of indicating a relative path (which is normal and real), somehow indicates an absolute path despite indicating the opposite, (c) allows dropping the initial / which *precludes* further extensions that would allow encoding abstract addresses.

Also, this completely misses the 3 unix socket types (STREAM/DGRAM/SEQPACKET). These map directly onto tcp/udp/{nothing}. What's the convention for these?

I'm implementing a libp2p unix-domain back-end where I need to parse Multiaddrs and I would like to not clash with the ecosystem. (If there is no ecosystem I will define one following the principles below.)

The two real address formats are:
1. normal (available universally): it's just a length-limited path (i.e. a sequence of non-0 bytes)
2. abstract (available under linux): it's a binary blob of length [1, 108], the first byte is 0

(1) are subject to normal path lookup (2) are used directly as keys into a kernel hashmap

Canonically, these are rendered (1) directly and (2) with an `@` replacing embedded 0s (incl. the one at the start, observe `ss --unix`; these do NOT end at the second NUL, it's just convention to terminate them there; you can (and it is done) embed NULs anywhere). This is not acceptable for a serialisation format since an @ is a valid byte in a path.

I don't really see a reason to not render these directly (with an optional percent-encoding step for display, sure), so
`/unix/%2Frun%2Fsystemd%2Fnotify` ‒ (1) absolute path
`/unix/notify` ‒ (1) relative path, equivalent to the one above for peers in `/run/systemd`
`/unix/%009e48095dfe839881%2Fbus%2Fsystemd-timesyn%2Fbus-api-timesync` ‒ (2)
are all valid addresses of UNIX-domain sockets taken from my system.

To actually use them you would strictly need to add `/stream` (connection-based protocol like TCP) or `/seqpacket` (TCP 2, Linux-only) and `/dgram` (connection-less like UDP) HOWEVER they occupy the same namespace, so unlike in TCP/UDP, you CANNOT have both
`/unix/%2Frun%2Fsystemd%2Fnotify/stream` &
`/unix/%2Frun%2Fsystemd%2Fnotify/dgram`
(this fails at bind/sendto/connect time).

nabijaczleweli:
God that's a lot of writing to say "i think we should just percent-encode the address instead of making it worse". Implementation of this proposal in https://github.com/libp2p/rust-libp2p/pull/6056/files#diff-ccb7f1d7de0deedbdbbcb4262ff70cd8658a1a0178fb3795e1b1a9dbbd00f919R363

achingbrain:
@nabijaczleweli thanks for your feedback and the additional use cases.

> If this only governs the serialised form:
> If this governs both the serialised form and opines on ORM/binary forms:

As noted in the OP, this only affects the string representation of a multiaddr.

> (a) this implies that an initial / is inserted if a path doesn't have one or (b) this implies that a relative path somehow magically becomes an absolute one

The assumption in this PR is that a relative path is only useful in the current context - if you try to interpret it starting from a different directory the path may not resolve or may resolve to an unexpected file which also seems like a significant footgun.

I can appreciate a light-touch garbage-in/garbage-out approach however so I am not opposed to dropping the initial `/` removal/insertion and just encoding whatever the value contains, though that complicates handling abstract socket paths as you point out.

> i think we should just percent-encode the address

This sounds reasonable.

> abstract (available under linux): it's a binary blob of length [1, 108], the first byte is 0
> This is not acceptable for a serialisation format since an @ is a valid byte in a path.

I think there are two ways forward here:

1. Declare this out of scope and say that `/unix` only refers to filesystem paths.  This way we can keep garbage-in/garbage-out encoding (e.g. support relative paths that may later make no sense).  To handle this properly it may need a new protocol to be defined since as you point out, any string serialization of the address could also be interpreted as a filesystem path - `/unix-abstract` or something, I don't have a strong opinion on the name.

2. An alternative might be to say:

    1. All filesystem paths must be absolute
    1. Filesystem paths start with `/` (which we will now not strip, as per the above)
    1. Abstract socket paths start with `@`
    1. All paths are percent-encoded

achingbrain:
My preference would be for option 2 since relative paths in multiaddrs seem to be too fragile to be useful.  What do you think @MarcoPolo?

nabijaczleweli:
Sorry, I don't think this makes sense. "This is different if you resolve it in a different context" is true of *all* paths/addresses, the point of an address is to tell *the specific process you're giving it to* how to call the peer you're thinking of. The "root" directory and the current directory are *exactly the same* in this model, and *both* are start-of-lookup inodes stored in the resolving process, configurable by the resolving process. Restricting this to just the root directory is strictly worse (and more ambiguous and less readable and less convention-following) since controlling the root directory is more convoluted and more disruptive for the resolving process than controlling the current directory.

You could analyse inet sockets the same way: on the same host, depending on the time, network namespace, routing table, &c., two different inet addresses can resolve to the same peer and two different peers can be reachable at the same address. The same more obviously applies to dns entries.

I think splitting unix into unix and unix-abstract makes sense (and this additionally cleans up the into of "unix works everywhere, unix-abstract works on linux" from "unix sometimes only works on linux, depending on the data") and GIGO percent-encoding them *both* makes the most sense.

nabijaczleweli:
Please cf. #179 for my proposed wording.

achingbrain:
> "This is different if you resolve it in a different context" is true of all paths/addresses

It's not really.  Something like `/ip4/123.123.123.123/tcp/1234` is unambiguous (assuming you can route to the host, etc).  When resolving: `/unix/foo` in a tree like:

```
/
├─ foo
├─ bar/
│  ├─ foo
├─ baz/
│  ├─ qux/
│      ├─ foo
```

..which `foo` you get depends entirely on where you start from.  Resolving `/unix/%2Ffoo` gets you the same `foo` no matter which directory is the current working directory for the resolving process, though I supposed you could be chrooted or something.

> controlling the root directory is more convoluted and more disruptive for the resolving process than controlling the current directory

It's possible, if the resolving process doesn't have access to the root or some intermediate directory(s).  Maybe then you'd need relative paths and accept that it's all GIGO.

Anyway, like I said, I think just percent-encoding the address and dropping the `/` removal/insertion is reasonable.

> I think splitting unix into unix and unix-abstract makes sense

Great, it sounds like you're in favour of option 1.  I'm fine with that.

I think we can make the amendments to `/unix` here, then add `/unix-abstract` (or whatever it ends up being called) as a separate PR.

I see you've added a bunch of other stuff to #179, I don't want simply being able to escape unix paths (the real win here) to get bogged down in discussion of other features.

nabijaczleweli:
I'm happy with the current wording, I can rebase #179 when this is merged.

PR #179, opened 2025-06-12
Specify /unix. Add /unix-abstract. Add /stream, /seqpacket, /dgram

nabijaczleweli:
## Summary
See https://github.com/multiformats/multiaddr/pull/174

## Before Merge
- [x] https://github.com/multiformats/multicodec/pull/382

### DIFF
--- a/protocols.csv
+++ b/protocols.csv
@@ -2,6 +2,9 @@ code,	size,	name,	comment
 4,	32,	ip4,
 6,	16,	tcp,
 273,	16,	udp,
+403,	0,	stream,
+404,	0,	seqpacket,
+405,	0,	dgram,
 33,	16,	dccp,
 41,	128,	ip6,
 42,	V,	ip6zone,	rfc4007 IPv6 zone
@@ -13,7 +16,8 @@ code,	size,	name,	comment
 132,	16,	sctp,
 301,	0,	udt,
 302,	0,	utp,
-400,	V,	unix,	Percent-encoded path to a Unix domain socket
+400,	V,	unix,	Percent-encoded path to a unix-domain socket
+401,	V,	unix-abstract,	Percent-encoded address of a linux abstract unix-domain socket
 421,	V,	p2p,	preferred over /ipfs
 421,	V,	ipfs,	backwards compatibility; equivalent to /p2p
 444,	96,	onion,
--- a/protocols/unix.md
+++ b/protocols/unix.md
@@ -1,6 +1,6 @@
 # `unix`
 
-This protocol encodes a Unix domain socket path to a resource. In the string
+This protocol encodes a UNIX-domain socket path to a resource. In the string
 representation, the path is encoded in a way consistent with a single URI Path
 segment per [RFC 3986 Section 3.3](https://datatracker.ietf.org/doc/html/rfc3986#autoid-23).
 
@@ -12,7 +12,7 @@ representation to avoid ambiguities over which characters were escaped.
 
 ## Examples
 
-The following is a table of examples converting some common Unix paths to their
+The following is a table of examples converting some common UNIX paths to their
 Multiaddr string form.
 
 | Unix Path                   | Multiaddr string form                   |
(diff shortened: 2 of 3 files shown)


rvagg:
I'm going to leave this and https://github.com/multiformats/multicodec/pull/382 up to @aschmahmann, @achingbrain et. al. to make a call on. Seems reasonable to me from a distance.

nabijaczleweli:
Rebased

rvagg:
ping @achingbrain @achingbrain, or perhaps @lidel?

protocols/unix.md
```diff
@@ -33,5 +33,29 @@ Multiaddr string form.
 appear anywhere, for example in the case where we route through some sort of
 proxy server or SSH tunnel.
 
-The absence of a `/` character at the start of the decoded address indicates a
-relative path, otherwise the path is absolute.
```
achingbrain, line 37: These lines may have been removed by accident?  They should probably remain in.

protocols/unix.md
```diff
@@ -1,6 +1,6 @@
 # `unix`
 
-This protocol encodes a Unix domain socket path to a resource. In the string
+This protocol encodes a UNIX-domain socket path to a resource. In the string
```
achingbrain: Small nit but [as I understand it](https://en.wikipedia.org/wiki/Unix) `UNIX` is a [trademark](https://unix.org/trademark.html) whereas `Unix` or `*nix` refers to generic or UNIX-like software - I don't think we're referring to any specific Unix here so this should probably remain as `Unix`?

achingbrain (commented):
LGTM - a couple of comments inline.

protocols/unix.md
nabijaczleweli: applied

protocols/unix.md
nabijaczleweli: applied

protocols.csv
```diff
@@ -2,6 +2,9 @@ code,	size,	name,	comment
 4,	32,	ip4,
 6,	16,	tcp,
 273,	16,	udp,
+403,	0,	stream,
+404,	0,	seqpacket,
+405,	0,	dgram,
```
lidel, line 7: @nabijaczleweli  these seem to be one-off.

https://github.com/multiformats/multicodec/pull/382 uses 402-404, this uses 403-405, need to be unified and i assume we dont want a gap so this PR should be 
```suggestion
402,	0,	stream,
403,	0,	seqpacket,
404,	0,	dgram,
```

?

protocols/unix.md
```diff
@@ -33,5 +36,29 @@ Multiaddr string form.
...
+
+| Rendered Address                             | multiaddr string form                                              |
+| -------------------------------------------- | ------------------------------------------------------------------ |
+| @f87a1c847a4ecaf3/bus/systemd/bus-api-system | `/unix-abstract/f87a1c847a4ecaf3%2Fbus%2Fsystemd%2Fbus-api-system` |
+| @/run/fsid.sock@@@@@@@@@@@@@@@@@@@@@@@@@@... | `/unix-abstract/%2Frun%2Ffsid.sock%00%00%00%00%00%00%00%00%00...`  |
+
+# `stream`, `seqpacket`, `dgram`
+
+These correspond to the *type* of Unix-domain socket:
+`SOCK_STREAM`, `SOCK_SEQPACKET`, `SOCK_DGRAM`.
```
lidel, line 60: Usage position unclear: Where do these appear in a multiaddr? After `/unix`? Example needed.

protocols/unix.md
```diff
@@ -33,5 +36,29 @@ Multiaddr string form.
...
+
+Previous versions of this specification did not contain these types;
+for compatibility, their absence should not indicate an error,
+if a default makes sense when decoding.
```
lidel, line 64: Default behavior underspecified: The doc says "their absence should not indicate an error, if a default makes sense when decoding" but doesn't specify what the default should be. Document that SOCK_STREAM should be assumed when no type is present?

This brings the other gap, use case not documented: Why would someone need to specify socket type at the multiaddr level? Most Unix socket usage is SOCK_STREAM. What's the practical scenario requiring SOCK_SEQPACKET or SOCK_DGRAM in a multiaddr?

protocols/unix.md
```diff
@@ -33,5 +36,29 @@ Multiaddr string form.
...
+In the following table, the address column follows the userspace convention
+of 0 bytes in the address being rendered as an `@` for abstract addresses
+for display only.
+
+| Rendered Address                             | multiaddr string form                                              |
+| -------------------------------------------- | ------------------------------------------------------------------ |
+| @f87a1c847a4ecaf3/bus/systemd/bus-api-system | `/unix-abstract/f87a1c847a4ecaf3%2Fbus%2Fsystemd%2Fbus-api-system` |
+| @/run/fsid.sock@@@@@@@@@@@@@@@@@@@@@@@@@@... | `/unix-abstract/%2Frun%2Ffsid.sock%00%00%00%00%00%00%00%00%00...`  |
+
+# `stream`, `seqpacket`, `dgram`
```
lidel, line 57: nit: these are generic terms exist outside of unix sockets (streams are very generic, datagrams also), would it be better to keep prefix `sock-` for all three?

protocols/unix.md
```diff
@@ -10,6 +10,9 @@ representation, no encoding is needed as the value is length prefixed.
 When comparing multiaddrs, implementations should compare their binary
 representation to avoid ambiguities over which characters were escaped.
 
+The absence of a `/` character at the start of the decoded address indicates a
```
lidel, line 13:
```suggestion
The absence of a `/` character (encoded as `%2F`) at the start of the decoded address indicates a
```

protocols/unix.md
```diff
@@ -33,5 +36,29 @@ Multiaddr string form.
...
 proxy server or SSH tunnel.
 
-The absence of a `/` character at the start of the decoded address indicates a
-relative path, otherwise the path is absolute.
+# `unix-abstract`
+
+This protocol encodes a Linux Unix-domain abstract socket address,
+which are distinguished by their first byte being 0.
+It is encoded the same way as `unix`;
+the marker byte is not part of the path.
```
lidel, line 44: The spec should have clear rationale why separate namespace is used instead of putting `%00` in `/unix`, because there will be people / LLMs that naively try to put `%00` in  `/unix`.

That being said,  @nabijaczleweli @achingbrain  apologies if this was discussed elsewhere, but why are we adding second namespace and not encode the first byte explicitly?

for example, instead 
- `/unix-abstract/f87a1c847a4ecaf3%2Fbus%2Fsystemd%2Fbus-api-system`

could we just prefix `%00`
- `/unix/%00f87a1c847a4ecaf3%2Fbus%2Fsystemd%2Fbus-api-system`

Any downside to doing this and removing `/unix-abstract`? (please document one way or another)

lidel (changes requested):
Thank you for improving these specs @nabijaczleweli

Before its ready,  looks like the codes are shifted + we need to add bit more details (defaults, rationale if/why we need separate namespace for abstract ones etc).

See details inline.

